PR #6, opened 2015-08-05
Added development guide for this plugin

recena:
The maintainer of this plugin (@stephenc) recommended to add this development guide.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,153 @@
+# Development Ideology
+
+Truths which we believe to be self-evident (adapted from [TextSecure's](https://github.com/WhisperSystems/TextSecure/blob/master/contributing.md))
+
+1. **The answer is not more options.** If you feel compelled to add a preference that's exposed to the user, it's very possible you've made a wrong turn somewhere.
+2. **There are no power users.** The idea that some users "understand" concepts better than others has proven to be, for the most part, false. If anything, "power users" are more dangerous than the rest, and we should avoid exposing dangerous functionality to them.
+3. **If it's "like PGP," it's wrong.** PGP is our guide for what not to do.
+4. **It's an asynchronous world.** We wary of anything that is anti-asynchronous: ACKs, protocol confirmations, or anly protocol-level "advisory" message.
+5. **There is no such thing as time**. Protocol ideas that require synchonized clocks are doomed to failure.
+
+# Code Style Guidelines
+
+## Resulting from long experience
+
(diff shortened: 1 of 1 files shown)


recena:
@reviewbybees Could you review these changes?

stephenc:
:bee:

stephenc:
:+1: as maintainer

CONTRIBUTING.md
```diff
@@ -0,0 +1,153 @@
+# Development Ideology
+
+Truths which we believe to be self-evident (adapted from [TextSecure's](https://github.com/WhisperSystems/TextSecure/blob/master/contributing.md))
+
+1. **The answer is not more options.** If you feel compelled to add a preference that's exposed to the user, it's very possible you've made a wrong turn somewhere.
+2. **There are no power users.** The idea that some users "understand" concepts better than others has proven to be, for the most part, false. If anything, "power users" are more dangerous than the rest, and we should avoid exposing dangerous functionality to them.
+3. **If it's "like PGP," it's wrong.** PGP is our guide for what not to do.
```
daniel-beck: Half of these don't even make sense for Jenkins plugins. I mean, sure, they're right, but this isn't TextSecure, so they shouldn't be front and center.

stephenc: They make sense to me in the context of Jenkins plugins. I removed the one that I felt did not make sense. Care to point out ones you feel could be rephrased to make more obvious?

daniel-beck: Jenkins isn't an instant messaging service, so the last two just aren't that relevant.

We're not doing a lot of user-controlled crypto either, so I'm not sure how relevant the PGP reference is.

The first two in the context of Jenkins are also debatable. Anything requiring `RunScripts` clearly is a power user feature and there are tons of them.

stephenc: 1. We get into a *lot* of trouble with all the myriad of interacting options that we expose to the user. By all means it makes sense to have plugable strategies which are extension points that users can inject into their instance by installing the plugin. But when you expose these options to *every* user then that is a usability failure. I want people to stop and think before they add just another checkbox.
2. If we need to add options "for power users" then we are doing something wrong again... we should be designing the UI so that features are easily accessible to all. The proliferation of `RunScripts` hacks point to a deeper issue.
3. If you are adding a CLI command and you make it's options like the PGP command lines...  (Not literally, but figuratively)

stephenc: As for the async... with workflow and even this plugin, we are dealing with separate JVMs on separate machines. If you have to coordinate anything between these then points 4 and 5 apply.

PR #4, opened 2015-07-16
[JENKINS-28689] Make compatible with Workflow plugin

recena:
Downstream of https://github.com/jenkinsci/workflow-plugin/pull/161

https://issues.jenkins-ci.org/browse/JENKINS-28689

- [x] Initial source code modifications
- [x] Implement a basic mocked SSH Server
- [x] Add a test to verify that SSH Agent is working how we expect
- [x] Add a test to verify that SSH Agent is working how we expect using a Workflow script
- [x] Implement `SSHAgentStep`, `DescriptorImpl` and `SSHAgentExecution`
- [x] Review the documentation
- [x] Add more tests
- [x] Snippet Generator must work

PR #7, opened 2015-08-06
Updated workflow version

recena:
This PR is related to https://github.com/jenkinsci/ssh-agent-plugin/pull/4


recena:
@reviewbybees Could you review these changes? Specially, @stephenc

jglick:
:bee: but I think the agreement is that no PR is needed for changes of this kind, as you are merely updating from a snapshot to the corresponding release.

PR #8, opened 2015-08-24
[FIXED JENKINS-30109] Update to bouncycastle 1.53

seryl:
https://issues.jenkins-ci.org/browse/JENKINS-30109

Fixes issues connecting to ECDH ssl servers /w jdk8, perhaps other fixes.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/mina/MinaRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/mina/MinaRemoteAgent.java
@@ -106,7 +106,6 @@ public class MinaRemoteAgent implements RemoteAgent {
             }
         } catch (Exception e) {
             e.printStackTrace(listener.error(Messages.SSHAgentBuildWrapper_UnableToReadKey(e.getMessage())));
-            e.printStackTrace(listener.getLogger());
         }
     }
 
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/mina/MinaRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/mina/MinaRemoteAgent.java
@@ -29,8 +29,11 @@ import com.cloudbees.jenkins.plugins.sshagent.RemoteAgent;
 import hudson.model.TaskListener;
 import org.apache.sshd.agent.unix.AgentServer;
 import org.apache.sshd.common.util.SecurityUtils;
-import org.bouncycastle.openssl.PEMReader;
-import org.bouncycastle.openssl.PasswordFinder;
+import org.bouncycastle.openssl.jcajce.JcePEMDecryptorProviderBuilder;
+import org.bouncycastle.openssl.jcajce.JcaPEMKeyConverter;
+import org.bouncycastle.openssl.PEMEncryptedKeyPair;
+import org.bouncycastle.openssl.PEMDecryptorProvider;
+import org.bouncycastle.openssl.PEMParser;
 
 import java.io.IOException;
 import java.io.StringReader;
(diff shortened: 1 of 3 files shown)


recena:
@seryl Could you add an Jira issue about this?

seryl:
https://issues.jenkins-ci.org/browse/JENKINS-30109

recena:
@seryl Thanks so much. I'll try your proposal.

recena:
@seryl If you are working on this ticket, it would be interesting you accept the Jira ticket.

seryl:
@recena I'm not sure what you mean?

All of the work is done to get it finalized. Jenkins core also needed to be updated to 152, and as far as I'm aware https://github.com/jenkinsci/instance-identity-module/pull/3 does that.

I can assign the tickets to myself in the Jenkins Jira, but they are already done and I don't have merge permissions; so I'm not sure how that helps..? Let me know what further steps I need to do to pull these in.

Happy to get this finished.

recena:
@seryl I've update [JENKINS-30109](https://issues.jenkins-ci.org/browse/JENKINS-30109) status.

recena:
@seryl I can merge this PR but I need to review it.

seryl:
@recena Any chance you could look at this?

36 additions.
21 deletions.

recena:
@seryl Could you squash the commits?

src/main/java/com/cloudbees/jenkins/plugins/sshagent/mina/MinaRemoteAgent.java
```diff
@@ -83,22 +86,27 @@ public void addIdentity(String privateKey, final String passphrase, String comme
...
+                } else if (o instanceof KeyPair) {
+                    keyPair = ((KeyPair) o);
                 }
+                agent.getAgent().addIdentity(keyPair, comment);
             } finally {
                 r.close();
             }
         } catch (Exception e) {
             e.printStackTrace(listener.error(Messages.SSHAgentBuildWrapper_UnableToReadKey(e.getMessage())));
+            e.printStackTrace(listener.getLogger());
```
recena: @seryl Why do you print `listener.getLogger()`?

seryl: I did it to match the mirrored configuration on JNRRemoteAgent.java

https://github.com/jenkinsci/ssh-agent-plugin/blob/master/src/main/java/com/cloudbees/jenkins/plugins/sshagent/jna/JNRRemoteAgent.java#L100

If this is something you want, I assume you want it in both places.
If not, I assume you don't want it in either.

seryl:
squashed and should be ready

recena:
@stephenc I agree with the changes. Your feedback will be great here.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/mina/MinaRemoteAgent.java
stephenc: They should both be using the `listener.error` form

stephenc:
Minor issue on logging. Would be great if a test could be included to prevent regressions. Otherwise looking good

seryl:
@stephenc I'm not sure what you mean... do you want me to validate that it's printing errors when they occur?

If that's the case, I'll just remove that. I have little to no interest in getting that logging information in, I just want the bouncycastle updates. I just did it since that seemed like what the initial intention was.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/mina/MinaRemoteAgent.java
seryl: Did we not want to print the getlogger then?

stephenc: `listener.error` also sends to `getLogger` but with the prefix message and allows for the two to be associated.

stephenc:
@seryl I want you to test that ECDH keys do not blow up. A test with a test resource of an ECDH key that gets set as a credential and passed through the MinaAgent would suffice. Testing the logging information is not required.

seryl:
I feel like there's some confusion..?

ECDH ssl is what we're adding support for here, not ecdsa keys.

We're allowing the key agreement when connecting to openssl servers to use the ecdh cipher.

This has nothing to do with openssh, so I don't understand why you are looking for a test on a new type of key?

This is to support downstream changes in the identity module.

Note: The original title was upgrade bouncycastle to 152; so that there would not be any confusion, which there seems to be.

Link to identity PR: https://github.com/jenkinsci/instance-identity-module/pull/3

When both of these are added, we'll be able to connect to apache and nginx servers which are using the openssl ciphers such as: ECDHE-ECDSA-AES256-GCM-SHA384.

The only change provided here for openssh, is to use the new bouncycastle interfaces, nothing more.

stephenc:
Ahh ok, so you just want to bump the BC version (with the usual hassle of method signature changes)

OK... I'd prefer then if you update the PR title to reflect that as otherwise when we document the changes users will get confused just as I did

seryl:
@stephenc Thanks for getting back to me, updated as requested

seryl:
@stephenc https://github.com/jenkinsci/instance-identity-module/pull/3 is the associated change, this should be the last of them to allow the upgrade to 152 for vanilla jenkins

TheSavior:
Just like https://github.com/jenkinsci/instance-identity-module/pull/3 this has been an issue for us as well and this fix works for us. Would be great to get this merged.

seryl:
@stephenc @recena Can we please give this some attention? It's been over 2 months now.

stephenc:
Still waiting for the requested changes to be made

seryl:
I'm not sure I follow. What requested changes?

stephenc:
You removed the correct way of logging the stack trace in one of the lines you changed. I requested you to revert that change in the PR comment on that line

seryl:
@stephenc removed as requested.

PR #9, opened 2015-12-04
[WIP] Add support for branch properties with branch-api based projects

stephenc:
@reviewbybees

We need a custom implementation for ssh-agent as there may be other ssh-agent build wrappers injected into the project that we are expected to merge into a single one...

I may consider an alternative approach whereby one and only one agent is started and the credentials all get merged into the existing one... which again will need support here that is over-and-above what you could deliver with a generic build wrapper support

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,11 @@
       <artifactId>ssh-credentials</artifactId>
       <version>1.11</version>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>branch-api</artifactId>
+      <version>1.0</version>
+    </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
     <dependency>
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.532.2</version>
+    <version>1.609.1</version>
```
recena: How it is possible? In `master` the baseline is 1.609.1

jglick:
I am not sure I follow the use case (and there is no description).

Presumably this is for `literate`, since decorating a `WorkflowMultiBranchProject` in this way is not possible. So basically you are saying: all branch projects for this repo, or at least most, should get the agent wrapper. Fine, but why would this not just be something configured in the literate repo?

Perhaps you just want to keep the credentials ID outside source control. But then why not just have a generic `BranchProperty` that lets you add any `BuildWrapper` of your choice? No need to add a `branch-api` dep to every plugin that happens to have a wrapper.

PR #10, opened 2016-02-29
There was one test without SSH Agent wrapper

recena:
@jglick noted it [here](https://github.com/jenkinsci/ssh-agent-plugin/pull/4/files#r54475090) in #4

@reviewbybees, specially @jglick


src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapperTest.java
```diff
@@ -106,6 +106,9 @@ public void sshAgentUnavailable() throws Exception {
 
         FreeStyleProject job = r.createFreeStyleProject();
 
+        SSHAgentBuildWrapper sshAgent = new SSHAgentBuildWrapper(credentialIds, false);
+        job.getBuildWrappersList().add(sshAgent);
+
```
jglick: But I thought the point of _this_ test was to show that the `ssh` command _would_ fail if you did not add the wrapper?

jglick:
:bee:

PR #11, opened 2016-02-29
[JENKINS-27152] Use a standardized directory for $SSH_AUTH_SOCK

jglick:
[JENKINS-27152](https://issues.jenkins-ci.org/browse/JENKINS-27152)

@reviewbybees


jglick:
Will check test failure; do not think I get it locally.

amuniz:
Looks good, but the test fail is suspicious.

The test passes for me locally (even with some stacktraces). I can see this log line:

```
[sshAgentAvailableAfterRestart #1] SSH Agent before restart /tmp/jenkins2298050772070608259.jnr
```

But the same line (guessed by position) in the CI build log is:

```
[sshAgentAvailableAfterRestart #1] SSH Agent before restart /tmp/ssh1266015111258068214
```

PR #14, opened 2016-06-15
[JENKINS-32120] Register Bouncy Castle on the remote agent by using Bouncy Castle API plugin

alvarolobato:
[JENKINS-32120](https://issues.jenkins-ci.org/browse/JENKINS-32120)

@reviewbybees esp @stephenc @oleg-nenashev @amuniz

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
@@ -24,10 +24,9 @@
 
 package com.cloudbees.jenkins.plugins.sshagent;
 
-import java.io.PrintStream;
-
 import javax.annotation.Nonnull;
 
+import hudson.model.TaskListener;
 import hudson.remoting.Channel;
 import hudson.remoting.VirtualChannel;
 import jenkins.bouncycastle.api.InstallBouncyCastleJCAProvider;
@@ -43,17 +42,16 @@ public class RemoteHelper {
      * @param channel to communicate with the agent
      * @param logger to log the messages
      */
-    public static void registerBouncyCastle(@Nonnull VirtualChannel channel, @Nonnull final PrintStream logger) {
+    public static void registerBouncyCastle(@Nonnull VirtualChannel channel, @Nonnull final TaskListener listener) {
         if (channel instanceof Channel) {
             try {
                 InstallBouncyCastleJCAProvider.on((Channel) channel);
-                logger.println("[ssh-agent] Registered BouncyCastle on the remote agent");
+                listener.getLogger().println("[ssh-agent] Registered BouncyCastle on the remote agent");
             } catch (Exception e) {
-                logger.println("[ssh-agent] WARNING: could not register BouncyCastle on the remote agent." + e.getMessage());
-                e.printStackTrace(logger);
+                e.printStackTrace(listener.error("[ssh-agent] WARNING: could not register BouncyCastle on the remote agent."));
             }
         } else {
-            logger.println("[ssh-agent] Skipped registering BouncyCastle, not running on a remote agent");
+            listener.getLogger().println("[ssh-agent] Skipped registering BouncyCastle, not running on a remote agent");
         }
     }
 }
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
@@ -26,6 +26,9 @@ package com.cloudbees.jenkins.plugins.sshagent;
 
 import javax.annotation.Nonnull;
 
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
 import hudson.model.TaskListener;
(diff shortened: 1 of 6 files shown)


src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
```diff
@@ -0,0 +1,59 @@
...
+     * Registers Bouncy Castle on a remote node logging the result.
+     * 
+     * @param channel to communicate with the agent
+     * @param logger to log the messages
+     */
+    public static void registerBouncyCastle(@Nonnull VirtualChannel channel, @Nonnull final PrintStream logger) {
+        if (channel instanceof Channel) {
+            try {
+                InstallBouncyCastleJCAProvider.on((Channel) channel);
+                logger.println("[ssh-agent] Registered BouncyCastle on the remote agent");
```
recena: Great log!

recena:
🐝

src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
```diff
@@ -0,0 +1,59 @@
...
+            } catch (Exception e) {
+                logger.println("[ssh-agent] WARNING: could not register BouncyCastle on the remote agent." + e.getMessage());
```
stephenc: if you passed a `TaskListener` this could be `e.printStackTrace(listener.error("[ssh-agent] WARNING: could not register BouncyCastle on the remote agent."));`

stephenc:
:ant: the helper method could have a more conventional signature using `TaskListener` which would allow for more flexible usage

alvarolobato:
@stephenc changed.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param logger to log the messages
+     */
+    public static void registerBouncyCastle(@Nonnull VirtualChannel channel, @Nonnull final TaskListener listener) {
+        if (channel instanceof Channel) {
+            try {
+                InstallBouncyCastleJCAProvider.on((Channel) channel);
+                listener.getLogger().println("[ssh-agent] Registered BouncyCastle on the remote agent");
+            } catch (Exception e) {
+                e.printStackTrace(listener.error("[ssh-agent] WARNING: could not register BouncyCastle on the remote agent."));
+            }
```
stephenc: :ant: no need for the `WARNING` as the logged line will get an `ERROR:` prefix anyway

stephenc:
:bee: (with one minor :ant:)

src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
recena: I prefer don't use `e.printStackTrace` when you can handle the behavior, print a user-friendly message and avoid an ugly trace in the UI.

stephenc: It may not be the prettiest but it will let you know what failed.

We do not expect it to fail, so it is not reasonable to expect us to anticipate all the potential failure modes and parse them (and test) so we can generate meaningful messages.

recena: As user I don't want to see a stack trace, simply something for understanding that it was wrong. As administrator or developer, I want these details but for that, exist the log files.

Different point of views, only that.

alvarolobato: I'm printing it because without it we are lost in order to know why it
didn't register. It is a real problem, we didn't manage it, just ignored it
because in some circumstances it has no consequences.

alvarolobato:
Removed "Warning"

alvarolobato:
@reviewbybees done

src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
```diff
@@ -0,0 +1,57 @@
...
+
+import hudson.model.TaskListener;
+import hudson.remoting.Channel;
+import hudson.remoting.VirtualChannel;
+import jenkins.bouncycastle.api.InstallBouncyCastleJCAProvider;
+
+/**
+ * Helper class for common remote tasks
+ */
+public class RemoteHelper {
```
amuniz: Maybe this class could be package visible only. This is currently public API usable from other plugins (and I fear others will use it).

No public API is not needed, in general.

amuniz:
:bee:

I would make the `RemoteHelper` class not visible outside though.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteHelper.java
alvarolobato: Not possible, being used from other two packages. I just added a `@Restricted`

amuniz:
re-:bee:

alvarolobato:
@reviewbybees done

PR #2, opened 2014-08-18
New implementation of ssh-agent

Mi-La:
I have created a new ssh-agent implementation. It is related to post here:
https://wiki.jenkins-ci.org/display/JENKINS/SSH+Agent+Plugin?focusedCommentId=73533285&#comment-73533285
This new implementation uses native programs available on a jenkins slave machine.
i.e. If ssh-agent program is available in PATH on a slave, it is used and identities are added using ssh-add.

The current ssh-agent-plugin implementation chooses a first supported ssh agent implementation. It would be greate if a user could assign priorities to ssh agent implementations in the plugin cofiguration.

PR #17, opened 2016-11-25
New implementation of ssh-agent (second attempt)

janvrany:
This is a second attempt to provide an alternative implementation of SSH agent by using
(remote) hosts native `ssh-agent`. 

It's a resurrection of earlier attempt made by @Mi-La some time ago - see [PR2][1]. The changes are following: 

 * @Mi-La's changes are rebased on top of current master
 * Code updated to catch up with API changes
 * Implemented @jglick suggestion made in original [PR2[[1] to use `Launcher` interface to check whether `ssh-agent` is installed. 

[1]: https://github.com/jenkinsci/ssh-agent-plugin/pull/2

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2012, CloudBees, Inc., Stephen Connolly.
+ * Copyright (c) 2014, Eccam s.r.o., Milan Kriz
  *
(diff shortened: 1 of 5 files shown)


lkraider:
It seems the test don't like the new ssh flag:
```
Bad protocol 2 host key algorithms '+ssh-dss'
```

janvrany:
Yeah, it does not. The question is how to solve this problem. I would have hoped CLI interface o older (open)ssh would remain compatible. Apparently not. So I need to hack tests again and only include the option iff client is >= 7.0. Sigh. 

There's another problem though - sshAgentAvailableAfterRestart is broken. This also needs further investigation...

jglick:
> This also needs further investigation...

Fixed that in #18 and am working on other cleanup and testing.

janvrany:
@jglick: Thank you so much for finishing this!

PR #19, opened 2017-03-06
JENKINS-42093 - fix quoting in askpass script

thomasgl-orange:
Change the askpass script from this:
```
#!/bin/sh
echo $SSH_PASSPHRASE
rm $0
```

...into that:
```
#!/bin/sh
echo "$SSH_PASSPHRASE"
rm "$0"
```

It fixes [JENKINS-42093](https://issues.jenkins-ci.org/browse/JENKINS-42093) (`rm` failing when there is a space in the workspace path), and allows using weird passphrases (with spaces, shell globs, etc.).


jamesj2:
This doesn't escape $ in the $SSH_PASSPHRASE. Wouldn't single quotes work better?

PR #20, opened 2017-05-17
JENKINS-44308 Avoid killing pre-existing ssh-agent

xverges:
Hide exsiting SSH_AGENT_PID from agents launched from the plugin

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgentFactory.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgentFactory.java
@@ -56,7 +56,7 @@ public class ExecRemoteAgentFactory extends RemoteAgentFactory {
     @Override
     public boolean isSupported(Launcher launcher, final TaskListener listener) {
         try {             
-            int status = launcher.launch().cmds("ssh-agent", "-k").quiet(true).start().joinWithTimeout(1, TimeUnit.MINUTES, listener);
+            int status = launcher.launch().cmds("ssh-agent", "-k").envs("SSH_AGENT_PID=").quiet(true).start().joinWithTimeout(1, TimeUnit.MINUTES, listener);
             /* 
              * `ssh-agent -k` returns 0 if terminates running agent or 1 if
              * it fails to terminate it. On Linux, 


src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgentFactory.java
```diff
@@ -56,7 +56,7 @@ public String getDisplayName() {
     @Override
     public boolean isSupported(Launcher launcher, final TaskListener listener) {
         try {             
-            int status = launcher.launch().cmds("ssh-agent", "-k").quiet(true).start().joinWithTimeout(1, TimeUnit.MINUTES, listener);
+            int status = launcher.launch().cmds("ssh-agent", "-k").envs("SSH_AGENT_PID=").quiet(true).start().joinWithTimeout(1, TimeUnit.MINUTES, listener);
             /* 
              * `ssh-agent -k` returns 0 if terminates running agent or 1 if
```
jglick: But then if there _were_ an existing agent, we would be returning 1 here (I suppose), meaning the wrapper _is_ supported, and then does this really work?

xverges:
> you are expected to run in an environment which has no agent to begin with

I was not aware of this restriction. I had to work in an environment where we used a pre-existing agent, since we did not have a shared SSH github key: Jenkins was started manually by a developer, after adding his key to a SSH agent.

Without this patch, after running a job relying on this plugin, subsequent checkouts would fail, since the pre-exitsing agent would be terminated. After the patch, we had no problem with our setup: checkouts worked and SSH commands using this plugin worked.

Regarding a test, what the patch does is clear an environment variable that on most environments would already be non set. What risks do you think the test should address?

jglick:
If you have specialized needs I would advise not using the plugin—just run `ssh-agent` yourself however you prefer.

xverges:
@jglick Jesse, sorry, I am confused by this discussion.
- `isSupported()` has the unintended side effect of unloading a key in some corner cases
- this unloading of a key made me spend several hours troubleshooting my build
- to save this effort to anyone that in the future shares my corner case, I've submitted a trivial patch that prevents this key from being unloaded unintentionally when checking if `ssh-agent` can be launched
- all the patch does is to unset an environment variable that the code assumes that is already unset

I honestly fail it see any risk, and I think it may help someone.

jglick:
Obsolete as of #152.

PR #21, opened 2017-08-25
plugin should have a description

jtnord:
add a description to the pom to avoid it showing up in the UC as the "jenkins parent pom" when installed.

oops

@reviewbybees


amuniz:
:bee:

alvarolobato:
:bee:

PR #23, opened 2018-01-11
Update plugin to make it runnable in PCT

oleg-nenashev:
- [x] Bump Jenkins Core to 2.7.3. The plugin was requiring Java 7 for 1.609.3 which was a bad pattern anyway
- [x] Bump Parent POM to the latest version, resolve upper bound conflicts
- [x] Bump BouncyCastle API dependency, just because there are collisions in PCT otherwise
- [x] Add test for JNRemoteAgent initialization, which proofs correct initialization with JEP-200 in PCT

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>bouncycastle-api</artifactId>
-      <version>1.0.2</version>
+      <version>2.16.2</version>
     </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.21</version>
+    <version>3.2</version>
     <relativePath />
   </parent>
 
@@ -55,10 +55,6 @@
     </developer>
   </developers>
 
-  <prerequisites>
-    <maven>2.2.1</maven>
-  </prerequisites>
-
   <scm>
     <connection>scm:git:git://github.com/jenkinsci/ssh-agent-plugin.git</connection>
     <developerConnection>scm:git:git@github.com:jenkinsci/ssh-agent-plugin.git</developerConnection>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -67,21 +63,21 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
-    <java.level>7</java.level> <!-- sshd-core is 7+ -->
+    <jenkins.version>2.7.3</jenkins.version>
```
oleg-nenashev: @stephenc I consider bumping it even more to 2.73 in order to pick SSHD Core 1.6.0 from the core instead of bundling the current 1.0.0. WDYT?

jglick: And removing `pluginFirstClassLoader`. Maybe, but a far riskier change, better left for another day.

pom.xml
```diff
@@ -100,13 +96,13 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
```
jglick: Delete, this gets inherited from the parent POM.

pom.xml
```diff
@@ -100,13 +96,13 @@
...
-      <version>1.7.7</version>
+      <version>1.7.25</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-jdk14</artifactId>
```
jglick: Ditto.

pom.xml
```diff
@@ -161,6 +157,13 @@
       <artifactId>workflow-cps</artifactId>
       <version>${workflow-jenkins-plugin.version}</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <!-- Provided by the core -->
```
jglick: Probably solved in a dep update FWIW.

src/test/java/com/cloudbees/jenkins/plugins/sshagent/jna/JNARemoteAgentTest.java
```diff
@@ -0,0 +1,61 @@
...
+import hudson.slaves.DumbSlave;
+import org.junit.Assume;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+/**
+ * Tests for {@link JNRRemoteAgent}.
+ */
+public class JNARemoteAgentTest {
```
jglick: `JNARemoteAgent`, as in the test name, or `JNRRemoteAgent`, as in the comment?

oleg-nenashev:
@reviewbybees done

oleg-nenashev:
I can address the comments above, bot only if the current maintainer (@stephenc?) is interested in the PR.

stephenc:
@oleg-nenashev if you want to address jesse's comments that would be great

jvz:
It'd be nice to still see this merged. This plugin has a lot of outdated dependencies.

oleg-nenashev:
Added it to my TODO list

jglick:
Conflicts with #26 I guess.

warden:
To be replaced by #30 and #31  ?

jglick:
Obsolete I think.

PR #24, opened 2018-06-26
avoid use of jnr-posix

ndeloof:
@jenkinsci/java10-support


ndeloof:
inactive, feel free to take over

PR #29, opened 2019-09-08
Do not use jnr-posix

jglick:
Refiling #24 since otherwise it would get lost.


warden:
Just a nitpick, I think it would be nicer to rebase instead of merging master to keep the history clean (I saw somewhere a git bot that does checks for the GH PRs automatically).

jglick:
> I think it would be nicer to rebase instead of merging master to keep the history clean

I do the exact opposite. Ultimately it is up to the maintainer of a repository whether they choose to squash-merge PRs and use destructive history operations in the origin repo. Doing so _within_ a PR branch seems particularly pointless, breaking some GH review functionality and losing track of CI status checks.

oleg-nenashev:
```
developers:
- "jglick"
- "kohsuke"
- "recena"
- "stephenconnolly"
```

PR #33, opened 2019-11-15
[JENKINS-59259] Fix durability crash in sshAgent step

jonsten:
This fixes https://issues.jenkins-ci.org/browse/JENKINS-59259. The bug occurs when the agent looses connection to the master during the step execution. Once the agent is reconnected the build is resumed and everything looks to be fine, but when the sshAgent step finishes, then an EOFException is thrown since the step/plugin tries to use write to a closed channel.

The root of the problem is that ExecRemoteAgent stored the launcher object. This is not safe in Jenkins Pipeline since the object changes when the agent is reconnected (or if the master is restarted). The fix is to retrieve the object on demand through StepContext. I've also addressed so that StepContext is used to retrieve other objects such as TaskListener, Workspace and the Run.

This is not the best PR I've made through my days, in the best of worlds the code should have been refactored, but my limited knowledge of the code base makes this hard. So, this is what I came up with, feed back is very much appreciated!

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteAgent.java
@@ -24,6 +24,9 @@
 
 package com.cloudbees.jenkins.plugins.sshagent;
 
+import hudson.Launcher;
+import hudson.model.TaskListener;
+
 import java.io.IOException;
 
 /**
@@ -43,12 +46,18 @@ public interface RemoteAgent {
      * @param privateKey the private key.
      * @param passphrase the passphrase or {@code null}.
      * @param comment    the comment to give to the key.
+     * @param launcher   the launcher for the remote node.
+     * @param listener   for logging.
      * @throws java.io.IOException if something went wrong.
      */
-    void addIdentity(String privateKey, String passphrase, String comment) throws IOException, InterruptedException;
+    void addIdentity(String privateKey, String passphrase, String comment, Launcher launcher,
+                     TaskListener listener) throws IOException, InterruptedException;
 
     /**
      * Stops the agent.
+     *
+     * @param launcher the launcher for the remote node.
+     * @param listener for logging.
      */
-    void stop() throws IOException, InterruptedException;
+    void stop(Launcher launcher, TaskListener listener) throws IOException, InterruptedException;
 }
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
@@ -305,6 +305,10 @@ public class SSHAgentBuildWrapper extends BuildWrapper {
          */
         private final RemoteAgent agent;
 
+        private final Launcher launcher;
+
+        private final BuildListener listener;
+
         /**
          * Construct the environment and initialize on the remote node.
          *
@@ -354,6 +358,8 @@ public class SSHAgentBuildWrapper extends BuildWrapper {
          */
         public SSHAgentEnvironment(Launcher launcher, BuildListener listener, @CheckForNull FilePath workspace) throws Throwable {
             RemoteAgent agent = null;
+            this.launcher = launcher;
+            this.listener = listener;
             listener.getLogger().println("[ssh-agent] Looking for ssh-agent implementation...");
             Map<String, Throwable> faults = new LinkedHashMap<String, Throwable>();
             for (RemoteAgentFactory factory : Jenkins.getActiveInstance().getExtensionList(RemoteAgentFactory.class)) {
@@ -395,7 +401,7 @@ public class SSHAgentBuildWrapper extends BuildWrapper {
             final Secret passphrase = key.getPassphrase();
             final String effectivePassphrase = passphrase == null ? null : passphrase.getPlainText();
             for (String privateKey : key.getPrivateKeys()) {
-                agent.addIdentity(privateKey, effectivePassphrase, description(key));
+                agent.addIdentity(privateKey, effectivePassphrase, description(key), launcher, listener);
             }
         }
 
@@ -414,7 +420,7 @@ public class SSHAgentBuildWrapper extends BuildWrapper {
         public boolean tearDown(AbstractBuild build, BuildListener listener)
                 throws IOException, InterruptedException {
             if (agent != null) {
-                agent.stop();
+                agent.stop(launcher, listener);
                 listener.getLogger().println(Messages.SSHAgentBuildWrapper_Stopped());
             }
             return true;
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
@@ -70,7 +70,7 @@ public class SSHAgentStepExecution extends AbstractStepExecutionImpl {
     @Override
     public void stop(Throwable cause) throws Exception {
         if (agent != null) {
-            agent.stop();
+            agent.stop(getContext().get(Launcher.class), listener);
             listener.getLogger().println(Messages.SSHAgentBuildWrapper_Stopped());
         }
         purgeSockets();
@@ -179,7 +179,8 @@ public class SSHAgentStepExecution extends AbstractStepExecutionImpl {
             final Secret passphrase = userPrivateKey.getPassphrase();
             final String effectivePassphrase = passphrase == null ? null : passphrase.getPlainText();
             for (String privateKey : userPrivateKey.getPrivateKeys()) {
-                agent.addIdentity(privateKey, effectivePassphrase, SSHAgentBuildWrapper.description(userPrivateKey));
+                agent.addIdentity(privateKey, effectivePassphrase, SSHAgentBuildWrapper.description(userPrivateKey),
+                        launcher, listener);
             }
         }
 
@@ -195,7 +196,7 @@ public class SSHAgentStepExecution extends AbstractStepExecutionImpl {
         try {
             TaskListener listener = getContext().get(TaskListener.class);
             if (agent != null) {
-                agent.stop();
+                agent.stop(getContext().get(Launcher.class), listener);
                 listener.getLogger().println(Messages.SSHAgentBuildWrapper_Stopped());
             }
         } finally {
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
@@ -44,13 +44,6 @@ public class ExecRemoteAgent implements RemoteAgent {
     private static final String AuthSocketVar = "SSH_AUTH_SOCK";
     private static final String AgentPidVar = "SSH_AGENT_PID";
     
-    private final Launcher launcher;
-    
-    /**
-     * The listener in case we need to report exceptions
-     */
-    private final TaskListener listener;
-
     private final FilePath temp;
     
     /**
@@ -62,15 +55,13 @@ public class ExecRemoteAgent implements RemoteAgent {
     private final Map<String, String> agentEnv;
 
     ExecRemoteAgent(Launcher launcher, TaskListener listener, FilePath temp) throws Exception {
-        this.launcher = launcher;
-        this.listener = listener;
         this.temp = temp;
 
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
         if (launcher.launch().cmds("ssh-agent").stdout(baos).start().joinWithTimeout(1, TimeUnit.MINUTES, listener) != 0) {
             throw new AbortException("Failed to run ssh-agent");
         }
-        agentEnv = parseAgentEnv(new String(baos.toByteArray(), StandardCharsets.US_ASCII)); // TODO could include local filenames, better to look up remote charset
+        agentEnv = parseAgentEnv(new String(baos.toByteArray(), StandardCharsets.US_ASCII), listener); // TODO could include local filenames, better to look up remote charset
         
         if (agentEnv.containsKey(AuthSocketVar)) {
             socket = agentEnv.get(AuthSocketVar);
@@ -91,7 +82,8 @@ public class ExecRemoteAgent implements RemoteAgent {
      * {@inheritDoc}
      */
     @Override
-    public void addIdentity(String privateKey, final String passphrase, String comment) throws IOException, InterruptedException {
+    public void addIdentity(String privateKey, final String passphrase, String comment, Launcher launcher,
+                            TaskListener listener) throws IOException, InterruptedException {
         FilePath keyFile = temp.createTextTempFile("private_key_", ".key", privateKey);
         try {
             keyFile.chmod(0600);
@@ -107,7 +99,7 @@ public class ExecRemoteAgent implements RemoteAgent {
                 }
                 
                 // as the next command is in quiet mode, we just add a message to the log
-                launcher.getListener().getLogger().println("Running ssh-add (command line suppressed)");
+                listener.getLogger().println("Running ssh-add (command line suppressed)");
                 
                 if (launcher.launch().quiet(true).cmds("ssh-add", keyFile.getRemote()).envs(env).stdout(listener).start().joinWithTimeout(1, TimeUnit.MINUTES, listener) != 0) {
                     throw new AbortException("Failed to run ssh-add");
@@ -126,7 +118,7 @@ public class ExecRemoteAgent implements RemoteAgent {
      * {@inheritDoc}
      */
     @Override
-    public void stop() throws IOException, InterruptedException {
+    public void stop(Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
         if (launcher.launch().cmds("ssh-agent", "-k").envs(agentEnv).stdout(listener).start().joinWithTimeout(1, TimeUnit.MINUTES, listener) != 0) {
             throw new AbortException("Failed to run ssh-agent -k");
         }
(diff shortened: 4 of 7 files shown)


src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
```diff
@@ -153,6 +154,49 @@ public void evaluate() throws Throwable {
...
 
+    /**
+     * This test verifies that sshAgent step handles that the build agent
+     * disconnects and reconnects during the step execution.
+     */
+    @Issue("JENKINS-59259")
+    @Test
+    public void agentConnectionDropTest() throws Exception {
+
+        story.addStep(new Statement() {
```
jglick: BTW if you use `then` you can pass a lambda as a body, which is a bit prettier.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteAgent.java
```diff
@@ -43,12 +45,15 @@
      * @param privateKey the private key.
      * @param passphrase the passphrase or {@code null}.
      * @param comment    the comment to give to the key.
+     * @param launcher the launcher for the remote node.
      * @throws java.io.IOException if something went wrong.
      */
-    void addIdentity(String privateKey, String passphrase, String comment) throws IOException, InterruptedException;
+    void addIdentity(String privateKey, String passphrase, String comment, Launcher launcher) throws IOException, InterruptedException;
```
jglick: Incompatible but I am inclined to let it pass given that [there are no _known_ external implementations](https://jenkins.io/doc/developer/extensions/ssh-agent/).

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -25,18 +26,6 @@
 
     private static final long serialVersionUID = 1L;
 
-    @StepContextParameter
```
jglick: Note [migration guide](https://javadoc.jenkins.io/plugin/workflow-step-api/org/jenkinsci/plugins/workflow/steps/AbstractStepExecutionImpl.html#AbstractStepExecutionImpl--) though this need not be done now (can be a follow-up).

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -83,8 +72,12 @@ public void onResume() {
             purgeSockets();
             initRemoteAgent();
         } catch (IOException | InterruptedException x) {
-            listener.getLogger().println(Messages.SSHAgentBuildWrapper_CouldNotStartAgent());
-            x.printStackTrace(listener.getLogger());
+            try {
+                getLogger().println(Messages.SSHAgentBuildWrapper_CouldNotStartAgent());
+                x.printStackTrace(getLogger());
```
jglick: `Functions.printStackTrace` preferred now BTW

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -229,4 +225,63 @@ private void purgeSockets() {
...
+
+    /**
+     * Retrieves the task listener reference in the current context.
+     *
+     * @return Current task listener object
+     *
+     * @throws IOException See {@link StepContext#get(Class)}
+     * @throws InterruptedException See {@link StepContext#get(Class)}
+     */
+    private TaskListener getListener() throws IOException, InterruptedException {
```
jglick: Tip: you can avoid `catch`ing these at call sites by falling back to either `TaskListener.NULL` or `LogTaskListener`.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
```diff
@@ -135,7 +132,7 @@ public void stop() throws IOException, InterruptedException {
     /**
      * Parses ssh-agent output.
      */
-    private Map<String,String> parseAgentEnv(String agentOutput) throws Exception{
+    private Map<String,String> parseAgentEnv(String agentOutput, TaskListener listener) throws Exception{
```
jglick: Why? Seems unused.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
```diff
@@ -44,8 +44,6 @@
     private static final String AuthSocketVar = "SSH_AUTH_SOCK";
     private static final String AgentPidVar = "SSH_AGENT_PID";
     
-    private final Launcher launcher;
-    
     /**
      * The listener in case we need to report exceptions
```
jglick: Probably better to pass the `TaskListener` as an argument to these methods too, so we do not need to rely on `Launcher.getListener()`.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
jonsten: It's used on line 140 and 144, or am I missing something?

src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
jonsten: Great tip, thanks!

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
jonsten: Did not know of that, changed.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
jonsten: I will follow up in separate PR.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteAgent.java
jonsten: My thought as well, not publicly used elsewhere and very unlikely that anyone have local implementations of this.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
jonsten: Never mind, adding it to this PR, seems like there is no good way of making dependent PRs on github.

jonsten:
@jglick, Thanks for the feedback! I've fixed all the things you pointed at, so this PR is ready for another review round.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
jglick: Sorry, saw a new method parameter without a corresponding body change and forgot that this was switching from a field reference to a parameter reference.

jglick:
From https://github.com/jenkinsci/ssh-agent-plugin/pull/33#discussion_r363639546

> seems like there is no good way of making dependent PRs on github

Actually there is, if you are comfortable with Git and have the inclination; I do this rather frequently. You just create a new branch based on the first one, make additional changes, and file that, with a description saying something like `Subsumes #nnn` for clarity and to make the backlink.

GitHub will display the aggregate changes in the second PR, so it will not at first be usefully reviewable (careless reviewers might actually carp about diff hunks coming from the first, subsumed PR). You can construct a permalink to display the incremental changes if you want (though they will not be reviewable); or you can even file a (draft) PR against your own fork displaying the aggregate changes (just select your first PR’s branch as the base), which will be fully reviewable. Usually it is not necessary to bother with such tricks, however: once the first PR has been merged, if you merge `origin/master` into your second branch¹ and push it, GitHub will display the incremental diff properly (it groks Git’s `...` semantics). Note that all these things get thrown out the window once somebody does destructive history operations, such as force-pushes or rebases or squash-merges; you need to use proper `merge`s!

----

¹My current favorite `~/.gitconfig` entry:

```ini
[alias]
  mergemaster = pull origin master:master
```

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
```diff
@@ -395,7 +398,7 @@ public void add(SSHUserPrivateKey key) throws IOException, InterruptedException
             final Secret passphrase = key.getPassphrase();
             final String effectivePassphrase = passphrase == null ? null : passphrase.getPlainText();
             for (String privateKey : key.getPrivateKeys()) {
-                agent.addIdentity(privateKey, effectivePassphrase, description(key));
+                agent.addIdentity(privateKey, effectivePassphrase, description(key), launcher, launcher.getListener());
```
jglick: Maybe better for `SSHAgentEnvironment` to hold onto a `TaskListener` too.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -56,6 +46,11 @@
      */
     private transient RemoteAgent agent = null;
 
+    @Inject
```
jglick:
```suggestion
```

Guiceless

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -262,26 +260,26 @@ private Launcher getLauncher() throws IOException, InterruptedException {
...
-     * @throws IOException See {@link StepContext#get(Class)}
-     * @throws InterruptedException See {@link StepContext#get(Class)}
      */
-    private TaskListener getListener() throws IOException, InterruptedException {
-        return getContext().get(TaskListener.class);
+    private TaskListener getListener() {
+        try {
+            return getContext().get(TaskListener.class);
+        } catch (IOException | InterruptedException e) {
+            return TaskListener.NULL;
```
jglick: Good idea to add a

```java
LOGGER.log(Level.WARNING, null, e);
```

here in case it is significant.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -25,18 +27,6 @@
...
-    private transient Run<?, ?> build;
-
-    @StepContextParameter
-    private transient Launcher launcher;
-
-    @StepContextParameter
-    private transient FilePath workspace;
-
     @Inject(optional = true)
     private SSHAgentStep step;
```
jglick: And get rid of this field too, and switch the supertype of `SSHAgentStep` from `AbstractStepImpl` to plain `Step`, so that its `start` method creates the `SSHAgentStepExecution` programmatically passing `this` (no Guice magic); the descriptor would also become plain `StepDescriptor`, so for `getRequiredContext` you use (say) `ImmutableSet.of` to list the context types you will need (`TaskListener` etc. as above). You should just be able to follow the deprecation warnings.

jonsten:
> From [#33 (comment)](https://github.com/jenkinsci/ssh-agent-plugin/pull/33#discussion_r363639546)
> 
> > seems like there is no good way of making dependent PRs on github
> 
> Actually there is, if you are comfortable with Git and have the inclination; I do this rather frequently. You just create a new branch based on the first one, make additional changes, and file that, with a description saying something like `Subsumes #nnn` for clarity and to make the backlink.
> 
> GitHub will display the aggregate changes in the second PR, so it will not at first be usefully reviewable (careless reviewers might actually carp about diff hunks coming from the first, subsumed PR). You can construct a permalink to display the incremental changes if you want (though they will not be reviewable); or you can even file a (draft) PR against your own fork displaying the aggregate changes (just select your first PR’s branch as the base), which will be fully reviewable. Usually it is not necessary to bother with such tricks, however: once the first PR has been merged, if you merge `origin/master` into your second branch¹ and push it, GitHub will display the incremental diff properly (it groks Git’s `...` semantics). Note that all these things get thrown out the window once somebody does destructive history operations, such as force-pushes or rebases or squash-merges; you need to use proper `merge`s!
> 
> ¹My current favorite `~/.gitconfig` entry:
> 
> ```ini
> [alias]
>   mergemaster = pull origin master:master
> ```

Thanks for taking the time to write this, appreciate it! Unfortunately it sort of sums up the solutions I've found thus far, i.e. it's not as seamless as I would like, either the reviewer needs to jump through hoops or the committer needs to do additional work. Anyways, let's leave it at that and focus on the code change :)

I will go through your comments and fix the code. As always, appreciate the feedback!

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
jonsten: I'm definitely not an expert, but is this a safe change, i.e. won't this lead to deserialization problems for builds that was ongoing during plugin upgrade?

jglick: > won't this lead to deserialization problems for builds that was ongoing during plugin upgrade

Normally this is why you use the constructor of `AbstractStepExecutionImpl` that is _not_ deprecated. In this case it is true that the `Step` instance is included in the serial state as well (that is not typical), so yes this would need to retain the original deprecated `AbstractStepImpl` supertype, yet `start` is `final`…hmm. Maybe better to deprecate `SSHAgentStep`, leave it untouched, but move its descriptor to a new `Step` implementation, and have `SSHAgentStepExecution.readResolve` switch fields.

All this is going to be kind of complicated, so better to not do it in this PR which is not really related. Even just switching `@StepContextParameter`s to `getContext().get(…)` is not quite safe, as that affects the computed mandatory context for the deprecated `AbstractStepDescriptorImpl`. Can we rip the conversion stuff out of this PR and keep it focused on removing `Launcher` (and, optionally, `TaskListener`) from the remoted state?

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStep.java
```diff
@@ -48,6 +51,11 @@ public SSHAgentStep(final List<String> credentials) {
         this.ignoreMissing = false;
     }
 
+    @Override
+    public StepExecution start(StepContext context) throws Exception {
+        return new SSHAgentStepExecution(this, context);
+    }
+
     @Extension
     public static final class DescriptorImpl extends AbstractStepDescriptorImpl {
```
jglick: This Guiceful class would need to be converted to nondeprecated `StepDescriptor`.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStep.java
```diff
@@ -21,7 +24,7 @@
 import java.util.Collections;
 import java.util.List;
 
-public class SSHAgentStep extends AbstractStepImpl implements Serializable {
+public class SSHAgentStep extends Step implements Serializable {
```
jglick: Unfortunately this will not be compatible for running builds.

jglick:
Moving to a top-level thread:

> Can we rip the conversion stuff out of this PR and keep it focused on removing `Launcher` (and, optionally, `TaskListener`) from the remoted state?

So this would mean reverting all changes to `SSHAgentStep`, and most changes to `SSHAgentStepExecution`: everything except adding the two new parameters to the `agent.stop` and `agent.addIdentity` calls. The (compatible) cleanup of deprecated Guice APIs can be deferred to a rainy day, and I might want to make `AbstractStepImpl.start` non`final` to make that easier anyway.

jonsten:
Once again, thanks for the feedback @jglick! I think I managed to remove all the Guice and refactoring related changes. The two commits should be the only essentials needed to fix the bug.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -17,24 +18,39 @@
...
 public class SSHAgentStepExecution extends AbstractStepExecutionImpl {
 
     private static final long serialVersionUID = 1L;
 
-    @StepContextParameter
+    /**
+     * @deprecated use {@link #getListener()} instead
+     */
+    @Deprecated
     private transient TaskListener listener;
```
jonsten: I'm not 100% sure about this part, so please correct me. As I understand it we need to leave the fields so that ongoing builds are deserialized correctly, but removing StepContextParameter can be removed as it is nolonger needed?

jglick: No. `transient` fields do not participate in deserialization. But the `@StepContextParameter` is inspected by the descriptor.

jglick:
> passed `, launcher, listener`

Or perhaps

```java
, getContext().get(Launcher.class), listener
```

since I suppose the `launcher` field will not have been updated in the circumstance covered by this bug. Uses of the `launcher` field during setup should be safe. and there is no need to touch the field itself.

jonsten:
Aha, I see, my bad. I'll fix.

jglick:
Thanks! I should be able to cut a release once this is ready.

jonsten:
Done. I've minimized the amount of changes to the execution class. I did however decide to keep most of the changes to the RemoteAgent implementations where I remove field references to launcher and listener. Reasoning behind this is that it easily could lead to incorrect usage where one reference old launchers/listeners. That is why e.g. RemoteAgent#addIndentity() still takes launcher and listener as argument, even though it only is used during startup.

Thank you for taking the time to review!

PR #30, opened 2019-09-20
[JENKINS-50181] Test and updated dependencies for changes in ssh-credentials for keys without a newline

warden:
[JENKINS-50181](https://issues.jenkins-ci.org/browse/JENKINS-50181)

PR #37, opened 2020-10-27
Use BOM and dependabot

warden:


pom.xml
```diff
@@ -29,12 +29,12 @@
...
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.24</version>
+    <version>4.12</version>
     <relativePath />
   </parent>
 
   <artifactId>ssh-agent</artifactId>
-  <version>1.21-SNAPSHOT</version>
+  <version>${revision}${changelist}</version>
```
jglick, line 37: Did you use `mvn incrementals:incrementalify`? You forgot to commit the `.mvn/` directory if so.

pom.xml
```diff
@@ -63,9 +63,11 @@
   </scm>
 
   <properties>
-    <jenkins.version>2.60.3</jenkins.version>
+    <revision>3.1.2</revision>
```
jglick, line 66: Copy-paste? Use the Maven tool.

```suggestion
    <revision>1.21</revision>
```

pom.xml
```diff
@@ -174,43 +166,23 @@
...
+        <type>pom</type>
+      </dependency>
       <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>structs</artifactId>
-        <version>1.14</version>
       </dependency>
       <dependency>
         <groupId>org.jenkins-ci</groupId>
         <artifactId>symbol-annotation</artifactId>
```
jglick, line 182: Delete this dep. It should be picked up from `structs`, never used as a direct dep.

pom.xml
```diff
@@ -174,43 +166,23 @@
...
   </dependencyManagement>
 
-  <build>
-    <plugins>
-      <plugin>
-        <groupId>org.jenkins-ci.tools</groupId>
-        <artifactId>maven-hpi-plugin</artifactId>
-        <extensions>true</extensions>
-        <configuration>
-          <pluginFirstClassLoader>true</pluginFirstClassLoader>
```
jglick, line 209: Revert, this is critical.

jglick (changes requested):
Several mistakes.

pom.xml
warden: Why is it critical (so I can put a comment above that it should not be touched) ?
Also, shouldn't below be ${java.level} instead of 1.5?

jglick: > Why is it critical

Because of one of the libraries it is using, SSHD IIRC. Without `pluginFirstClassLoader` that mode of the plugin will just throw errors. https://www.jenkins.io/doc/developer/plugin-development/dependencies-and-class-loading/

`compatibleSinceVersion` should also stay, though at this point it is mostly of historical interest since it is unlikely people would be updating since such an old version.

pom.xml
```diff
@@ -208,6 +185,7 @@
         <configuration>
           <pluginFirstClassLoader>true</pluginFirstClassLoader>
           <compatibleSinceVersion>1.5</compatibleSinceVersion>
+          <minimumJavaVersion>${java.level}</minimumJavaVersion>
```
jglick, line 188:
```suggestion
```

Nobody does this. https://github.com/jenkinsci/plugin-pom/pull/133 may do this automatically some day.

warden: This is added for compliancy with XSD as suggested by my intelliJ as a required param in `<configuration>`.

jglick: Well tell your IDE to shut up. :grin:

jglick: https://github.com/jenkinsci/plugin-pom/blob/0c6a61e618eaa913bd0c8489d12a58c78f0bdfed/pom.xml#L644 already defines this, so IDEA is presumably just not taking inheritance into account.

.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    ignore:
+      # core is using lower Guava version, don't touch it unless you want things to blow up on client installations at runtime
+      - dependency-name: "com.google.guava:guava"
```
jglick, line 9: Why is this here? This plugin’s `pom.xml` makes no reference to Guava. I think this whole `ignore` section can be deleted.

pom.xml
```diff
@@ -125,43 +124,36 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-api</artifactId>
```
jglick, line 126: Could be deleted; would be picked up transitively anyway.

pom.xml
```diff
@@ -125,43 +124,36 @@
...
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-cps</artifactId>
-      <version>2.45</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-support</artifactId>
```
jglick, line 151: ditto

pom.xml
```diff
@@ -125,43 +124,36 @@
...
-      <version>${workflow-support-plugin.version}</version>
       <scope>test</scope>
      </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-support</artifactId>
-      <version>${workflow-support-plugin.version}</version>
       <classifier>tests</classifier>
```
jglick, line 157: (but this needs to stay, if we are in fact using these utilities)

jglick (approved):

pom.xml
```diff
@@ -121,36 +121,16 @@
...
-      <scope>test</scope>
-    </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-job</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-basic-steps</artifactId>
```
jglick, line 139: You probably did need this.

pom.xml
```diff
@@ -121,36 +121,16 @@
...
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-durable-task-step</artifactId>
```
jglick, line 145: And this.

(Unless both are getting picked up from `docker-workflow`?)

warden: Tests are compiling fine and passing, why would I need this then?

jglick: Probably it is getting picked up from `docker-workflow` then.

warden:
@jglick thanks for help on this.  What's next? 

Iis the `<maintainers>` pom field still used across plugins and up2date in this case?  Looking at the merge history looking at the people who merged, it looks to me that there is a de-sync between maintainers and actual people that have repo merge rights.
After all, CI passed and there is an approval from you (so one of the people having merge rights) - wouldn't that work if in such case an auto-merge is done by the pipeline?

As a contributor, I'm just wondering if we can speed up the PR merging process somehow as it's not clear what happens here next.

jglick:
POM `maintainers` can be ignored. Maybe @MRamonLeon is maintaining these days?

warden:
@MRamonLeon any thoughts?

warden:
@MRamonLeon, @jglick ping again 🙏

MRamonLeon (approved):

MRamonLeon:
I don't see any reason to release these changes. Do you need it for any reason? Thank you for your contribution @warden

warden:
@MRamonLeon thanks for acting on this. No, this one is not needed to be released, more important is #30.

MRamonLeon:
I've commented on #30 @warden

pom.xml
jglick: Still missing `.mvn/extensions.xml`.

rantoniuk: Feel free to add that one file instead of commenting 2y old already merged PR ;-)
At the same time, might be just as well good to consider adding this file globally to all repositories under /jenkinsci org - I don't think it would harm them even if they're not using incrementals at the moment?

jglick: > Feel free to add that one file

Already done in #85.

> consider adding this file globally to all repositories

Could, but probably more confusing than helpful, since `mvn incrementals:incrementalify` expects to add it.

PR #18, opened 2017-02-08
[JENKINS-36997] CLI implementation of RemoteAgent

jglick:
[JENKINS-36997](https://issues.jenkins-ci.org/browse/JENKINS-36997); trying to finish #17.

- [X] passing tests
- [x] remove need for tests to pass `-o HostKeyAlgorithms=ssh-dss` ([reference](https://www.openssh.com/legacy.html))
- [X] normalize `ExecRemoteAgent.agent`
- [x] verify in https://github.com/jenkinsci/acceptance-test-harness/pull/272

@reviewbybees

PR #48, opened 2021-03-10
Delete non-exec-based agent factories

jglick:
After #18 there is not much reason to continue to support the in-JVM modes, and they are a maintenance burden, so deleting them.


alecharp:
I believe this is the way to go. I wonder if some kind of migration is required for this. What do you think @jglick ?

jglick:
I am not sure what migration is possible. As before, if you have the `ssh-agent` binary installed on your agent, the plugin will work. If you do not, we cannot do anything other than print an error message, which I think it already does (though as noted in the TODO comment, we could probably tune the logging behavior now).

alecharp:
I was thinking about some other plugin to use, other connection details to prefer.

jglick:
> some other plugin to use

I did note in docs the simpler alternative using `withCredentials` applicable to many use cases.

alecharp (approved):

pom.xml
```diff
@@ -152,13 +132,10 @@
...
     <plugins>
       <plugin>
-        <!-- critical because sshd needs to be loaded first, otherwise will just bail out -->
-        <!-- https://www.jenkins.io/doc/developer/plugin-development/dependencies-and-class-loading/ -->
         <groupId>org.jenkins-ci.tools</groupId>
         <artifactId>maven-hpi-plugin</artifactId>
         <extensions>true</extensions>
         <configuration>
-          <pluginFirstClassLoader>true</pluginFirstClassLoader>
           <compatibleSinceVersion>1.5</compatibleSinceVersion>
```
daniel-beck, line 139: Probably should have had this updated?

alecharp: we can for the next release. from https://github.com/jenkinsci/ssh-agent-plugin/pull/48#issuecomment-805759155, it shouldn't matter for `1.22` no?

jglick: IIUC `compatibleSinceVersion` is only relevant for settings changes, and there are none here.

daniel-beck: Doesn't this break existing configurations that rely on this feautre? It's not _exactly_ configuration compatibility in the strict sense, but I don't think anyone whose jobs we break will care about that distinction.

jglick: Whatever you think appropriate.

alecharp: If we have an agreement here, I need to cut a release really soon because I already published `1.22`

daniel-beck:
This change is breaking today's core security release on release.ci 🎉

PR #57, opened 2021-06-10
Update README.md - Jenkinsfile Example - With keyscan to prevent "Host key verification failed" errors

mcsage:
Added Jenkinsfile example to README - With keyscan to prevent "Host key verification failed" errors

- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,18 @@ And then your build will have those credentials available, e.g.
 
 From a Pipeline job, use the `sshagent` step.
 
+# Jenkinsfile
+
+```
+      steps {
+          sshagent (credentials: ['ssh-credentials-id']) {
+            sh 'mkdir ~/.ssh && chmod 0700 ~/.ssh'
+            sh 'echo "$(ssh-keyscan -t rsa,dsa example.com)" >> ~/.ssh/known_hosts'
+            sh 'ssh user@example.com ...'
+          }
+      }
+```
+
 # Version History
 
 For new versions, see [GitHub releases](https://github.com/jenkinsci/ssh-agent-plugin/releases).
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ From a Pipeline job, use the `sshagent` step.
 
 ```
       steps {
-          sshagent (credentials: ['ssh-credentials-id']) {
+          sshagent(credentials: ['ssh-credentials-id']) {
             sh '[ -d ~/.ssh ] || mkdir ~/.ssh && chmod 0700 ~/.ssh'
             sh 'echo "$(ssh-keyscan -t rsa,dsa example.com)" >> ~/.ssh/known_hosts'
             sh 'ssh user@example.com ...'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,18 @@ And then your build will have those credentials available, e.g.
 
 From a Pipeline job, use the `sshagent` step.
 
+# Jenkinsfile
+
+```
+      steps {
+          sshagent (credentials: ['ssh-credentials-id']) {
```
jglick, line 39:
```suggestion
          sshagent(credentials: ['ssh-credentials-id']) {
```

canonically

README.md
```diff
@@ -32,6 +32,18 @@ And then your build will have those credentials available, e.g.
...
+            sh '[ -d ~/.ssh ] || mkdir ~/.ssh && chmod 0700 ~/.ssh'
+            sh 'echo "$(ssh-keyscan -t rsa,dsa example.com)" >> ~/.ssh/known_hosts'
+            sh 'ssh user@example.com ...'
```
jglick, line 42:
```suggestion
            sh '''
                [ -d ~/.ssh ] || mkdir ~/.ssh && chmod 0700 ~/.ssh
                ssh-keyscan -t rsa,dsa example.com >> ~/.ssh/known_hosts
                ssh user@example.com ...
            '''
```

?

README.md
```diff
@@ -32,6 +32,18 @@ And then your build will have those credentials available, e.g.
 
 From a Pipeline job, use the `sshagent` step.
 
+# Jenkinsfile
+
+```
+      steps {
+          sshagent (credentials: ['ssh-credentials-id']) {
+            sh '[ -d ~/.ssh ] || mkdir ~/.ssh && chmod 0700 ~/.ssh'
+            sh 'echo "$(ssh-keyscan -t rsa,dsa example.com)" >> ~/.ssh/known_hosts'
```
jglick, line 41: Is this something that ought to be a direct feature of the plugin?

jglick: compare https://github.com/jenkinsci/ssh-agent-plugin/blob/11a5b8835f16807a81d63bbf24065f18d1f6256b/src/main/resources/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStep/help.html#L4

mcsage: Hi @jglick, yes indeed -  instead of setting StrictHostKeyChecking=no the plugin should get an feature to manage/automatically import hosts keys to known_hosts file/store.

jglick (approved):

README.md
joel-schaal: > `sh 'echo "$(ssh-keyscan -t rsa,dsa example.com)" >> ~/.ssh/known_hosts'`
This is actually circumventing the actual host verification: if there is a man-the-middle, `ssh-keyscan -t rsa,dsa example.com` will actually return that machine's fingerprint instead of the one from example.com.
If we already have the appropriate entry in known_hosts, this is mitigated by the fact that ssh will complain about multiple entries, but still it is a bit of a bad workaround.
What is actually needed, is to provide a static reference of the official fingerprints (config file or something similar, ideally)

jglick: You could hard-code such fingerprints, or use the `config-file-provider` plugin, etc.

Feel free to file an amending PR explaining what you think is a safer practice.

PR #127, opened 2023-04-28
Fix flaky test

jtnord:
observed several failures due to

```
java.net.BindException: Address already in use
	at java.base/sun.nio.ch.Net.bind0(Native Method)
	at java.base/sun.nio.ch.Net.bind(Net.java:459)
	at java.base/sun.nio.ch.Net.bind(Net.java:448)
	at java.base/sun.nio.ch.AsynchronousServerSocketChannelImpl.bind(AsynchronousServerSocketChannelImpl.java:164)
	at org.apache.sshd.common.io.nio2.Nio2Acceptor.bind(Nio2Acceptor.java:83)
	at org.apache.sshd.common.io.nio2.Nio2Acceptor.bind(Nio2Acceptor.java:173)
	at org.apache.sshd.server.SshServer.start(SshServer.java:331)
	at com.cloudbees.jenkins.plugins.sshagent.SSHAgentBase.startMockSSHServer(SSHAgentBase.java:171)
```

rather than obtain a port ant t1 and then use it later at t2 just get the SSH server to use a port and then find out which port is actually being used.

- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBase.java
+++ b/src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBase.java
@@ -8,14 +8,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.StringReader;
-import java.net.BindException;
-import java.net.InetAddress;
-import java.net.ServerSocket;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Paths;
 import java.security.PublicKey;
 import java.security.interfaces.RSAPublicKey;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Logger;
 
 import org.apache.sshd.common.config.keys.KeyUtils;
@@ -28,18 +24,19 @@ import org.apache.sshd.server.auth.pubkey.PublickeyAuthenticator;
 import org.apache.sshd.server.keyprovider.SimpleGeneratorHostKeyProvider;
 import org.apache.sshd.server.session.ServerSession;
 import org.apache.sshd.server.shell.ShellFactory;
+import org.junit.Rule;
+import org.junit.rules.TemporaryFolder;
 
 public class SSHAgentBase {
 
-    public static AtomicInteger SSH_SERVER_INITIAL_PORT = new AtomicInteger(4380);
+    @Rule
+    public TemporaryFolder temp = new TemporaryFolder();
 
     public static String SSH_SERVER_HOST = "localhost";
 
     public static String CREDENTIAL_ID = "84822271-02d5-47b8-b8ff-c40fef175c29";
 
     private SshServer sshd = null;
-
-    private int assignedPort = SSH_SERVER_INITIAL_PORT.getAndIncrement();
 
     public String KEY_WITHOUT_PASSWORD_AND_NO_NEWLINE =
             "-----BEGIN OPENSSH PRIVATE KEY-----\n" +
@@ -71,10 +68,9 @@ public class SSHAgentBase {
             "-----END OPENSSH PRIVATE KEY-----";
 
     protected void startMockSSHServer() throws Exception {
-        File hostKey = new File(System.getProperty("java.io.tmpdir") + "/key.ser");
-        hostKey.delete(); // do not carry from test to test
+        File hostKey = temp.newFile();
         sshd = SshServer.setUpDefaultServer();
-        sshd.setPort(getValidPort());
+        sshd.setPort(0);
         sshd.setHost(SSH_SERVER_HOST);
         SimpleGeneratorHostKeyProvider hostKeyProvider = new SimpleGeneratorHostKeyProvider(Paths.get(hostKey.getPath()));
         hostKeyProvider.setAlgorithm(KeyUtils.RSA_ALGORITHM);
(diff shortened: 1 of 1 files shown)


src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBase.java
```diff
@@ -28,19 +24,20 @@
...
 import org.apache.sshd.server.session.ServerSession;
 import org.apache.sshd.server.shell.ShellFactory;
+import org.junit.Rule;
+import org.junit.rules.TemporaryFolder;
 
 public class SSHAgentBase {
 
-    public static AtomicInteger SSH_SERVER_INITIAL_PORT = new AtomicInteger(4380);
+    @Rule
+    public TemporaryFolder temp = new TemporaryFolder();
```
jtnord, line 33: the same file was being used, which if using mutliple tests in parallel could lead to some funky timing issues also.

src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBase.java
```diff
@@ -71,10 +68,9 @@ public class SSHAgentBase {
             "-----END OPENSSH PRIVATE KEY-----";
 
     protected void startMockSSHServer() throws Exception {
-        File hostKey = new File(System.getProperty("java.io.tmpdir") + "/key.ser");
-        hostKey.delete(); // do not carry from test to test
+        File hostKey = temp.newFile();
         sshd = SshServer.setUpDefaultServer();
-        sshd.setPort(getValidPort());
+        sshd.setPort(0);
```
jtnord, line 73: get the SSH server to start on a random port - so that it is guaranteed to be available.

src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBase.java
```diff
@@ -313,36 +309,13 @@ public String getAuthorizedPublicKey() {
                 "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDXRix6tfUq1K3IbgXYMqKYdgwYU8hxgUiBwDtBpxq7smPT5s5xCg7C6vdYIitDfgLORbWheyCuHgOErwyoh2GAm842di5BjJVsw+WgxLUcM0gkUM06PJdM7I9hBuz8sxZM7Kcose382QSNWwg5cTQFhLzZr8ig5OtSYc6PPAjzR4GiX93akoZLcFaMmXZba8Qdqhr+zsrmCtgfxqAIJTXzVxtz0MLscTDk37ZtnRVnKXOOvy/a7hZumgKvTom+mSv0kq6hVNlUv2Q9jmNHp/RfLU5xc3aN0BninRXx7yaTfqXwhCZHfgbhG+B8PpMWZvwUBDsakxsVyGdEeGQo+n03 testkey-no-passphrase";
     }
 
-    /**
-     * Returns a valid port number to use in Mocked SSH Server. Verifies that the port is not being used.
-     *
-     * @return int with a valid port number
-     */
-    private int getValidPort() throws IOException {
```
jtnord, line 321: the port may be free immediately after it was obtained - but may have been used by something else before it was used by the SSH server.

src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBase.java
```diff
@@ -313,36 +309,13 @@ public String getAuthorizedPublicKey() {
...
-    }
-
     /**
      * Returns the assigned port number to be used in Mock.
      *
      * @return int with an assigned port number
      */
     public int getAssignedPort() {
-        return assignedPort;
+        return sshd.getPort();
```
jtnord, line 318: ask the ssh server for its port - it knows what it is!

src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBase.java
rsandell: Will JUnit actually apply this rule since it isn't a test suite?

jtnord: its a base class of the other tests so it should.  the test pass :)

rsandell: Meaning will the temp files created with it actually be cleaned up?

jtnord: yup!
Note that it was not cleaned up prior to this - one would always dangle.

src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBase.java
```diff
@@ -71,10 +68,9 @@ public class SSHAgentBase {
             "-----END OPENSSH PRIVATE KEY-----";
 
     protected void startMockSSHServer() throws Exception {
-        File hostKey = new File(System.getProperty("java.io.tmpdir") + "/key.ser");
-        hostKey.delete(); // do not carry from test to test
```
jtnord, line 75: because this was using the same file accross tests and deleting it if it existed!
(running multiple tests in parallel!)

rsandell (approved):

PR #144, opened 2024-04-04
Fail the entire build when a requested credential cannot be found

nattofriends:
[JENKINS-32104](https://issues.jenkins.io/browse/JENKINS-32104)

This is my first contribution to any public Jenkins code, hopefully didn't miss anything.

I also edited the message to interpolate the name of the requested credential; Credential Binding's message also does this.

### Testing done

- Added SSHAgentStepWorkflowTest.testMissingCredential

```[tasklist]
### Submitter checklist
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
@@ -138,7 +138,7 @@ public class SSHAgentStepExecution extends AbstractStepExecutionImpl implements 
             final SSHUserPrivateKey c = CredentialsProvider.findCredentialById(id, SSHUserPrivateKey.class, build);
             CredentialsProvider.track(build, c);
             if (c == null && !step.isIgnoreMissing()) {
-                listener.fatalError(Messages.SSHAgentBuildWrapper_CredentialsNotFound());
+                throw new IOException(Messages.SSHAgentBuildWrapper_CredentialsNotFound(id));
             }
             if (c != null && !userPrivateKeys.contains(c)) {
                 userPrivateKeys.add(c);
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
@@ -5,6 +5,7 @@ import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.google.inject.Inject;
 import edu.umd.cs.findbugs.annotations.CheckReturnValue;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.AbortException;
 import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Launcher;
@@ -138,7 +139,7 @@ public class SSHAgentStepExecution extends AbstractStepExecutionImpl implements 
             final SSHUserPrivateKey c = CredentialsProvider.findCredentialById(id, SSHUserPrivateKey.class, build);
             CredentialsProvider.track(build, c);
             if (c == null && !step.isIgnoreMissing()) {
-                throw new IOException(Messages.SSHAgentBuildWrapper_CredentialsNotFound(id));
+                throw new AbortException(Messages.SSHAgentBuildWrapper_CredentialsNotFound(id));
             }
             if (c != null && !userPrivateKeys.contains(c)) {
                 userPrivateKeys.add(c);
--- a/src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
+++ b/src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
@@ -8,6 +8,7 @@ import com.cloudbees.plugins.credentials.SystemCredentialsProvider;
 import hudson.Functions;
 import hudson.Launcher;
 import hudson.model.Fingerprint;
+import hudson.model.Result;
 import hudson.slaves.DumbSlave;
 import hudson.util.StreamTaskListener;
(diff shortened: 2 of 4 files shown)


src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -138,7 +138,7 @@ private void initRemoteAgent() throws IOException, InterruptedException {
             final SSHUserPrivateKey c = CredentialsProvider.findCredentialById(id, SSHUserPrivateKey.class, build);
             CredentialsProvider.track(build, c);
             if (c == null && !step.isIgnoreMissing()) {
-                listener.fatalError(Messages.SSHAgentBuildWrapper_CredentialsNotFound());
+                throw new IOException(Messages.SSHAgentBuildWrapper_CredentialsNotFound(id));
```
jglick, line 141: `AbortException` would be preferable, as mentioned in Jira, since it avoids printing a stack trace which in this context is not meaningful.

src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
```diff
@@ -282,4 +283,24 @@ public void sshAgentDocker() throws Exception {
         });
     }
 
+    @Issue("JENKINS-32104")
+    @Test
+    public void testMissingCredential() {
+        assumeFalse(Functions.isWindows());
+
```
jglick, line 290:
```suggestion
```
Does not seem necessary in this case, since the test does not rely on Linux specifics like `sh`.

src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
```diff
@@ -282,4 +283,24 @@ public void sshAgentDocker() throws Exception {
...
+            @Override
+            public void evaluate() throws Throwable {
+                WorkflowJob job = story.j.jenkins.createProject(WorkflowJob.class, "sshAgentAvailable");
+                job.setDefinition(new CpsFlowDefinition(""
+                  + "node {\n"
+                  + "  sshagent (credentials: ['nonexistent']) {\n"
+                  + "  }\n"
+                  + "}\n", true)
+                );
+                story.j.assertBuildStatus(Result.FAILURE, job.scheduleBuild2(0));
```
jglick, line 301: For “extra credit”, use `assertLogContains` to check for the expected error message. (Otherwise the test could pass for all sorts of incorrect reasons, such as a typo in the Pipeline script resulting in a Groovy syntax error!)

src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
```diff
@@ -282,4 +283,24 @@ public void sshAgentDocker() throws Exception {
         });
     }
 
+    @Issue("JENKINS-32104")
+    @Test
+    public void testMissingCredential() {
```
jglick, line 288: More extra credit if you are interested: test coverage for `ignoreMissing`, which seems to have been forgotten in #4 along with the incorrect handling of the default case.

jglick (approved):
Looks good, thanks! I left some minor suggestions (nothing blocking merge, if you are out of time).

I marked the issue **In Review** but could not find a Jira identifier for you to assign it to.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
nattofriends: I think I did this even though you did mention `AbortException` in Jira since https://github.com/jenkinsci/ssh-agent-plugin/blob/master/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java#L207 already exists, is it better to have them raise the same thing or to just raise `AbortException` here?

jglick: You can correct `SSHAgentBuildWrapper` (used for non-Pipeline projects) if you like, or leave it alone, so long as the Pipeline step does the right thing.

jglick (approved):

torbendury:
@nattofriends This bug was old enough for me to never notice it was actually buggy behavior :-)

Do I understand the new flag `ignoreMissing` correctly that it is going to "restore" the old behavior of pipeline scripts not failing completely when `sshagent` is unable to find credentials?

nattofriends:
> @nattofriends This bug was old enough for me to never notice it was actually buggy behavior :-)
> 
> Do I understand the new flag `ignoreMissing` correctly that it is going to "restore" the old behavior of pipeline scripts not failing completely when `sshagent` is unable to find credentials?

That flag isn't new, it has always been there. But my understanding is that it can be used to achieve the behavior you are looking for.

deribaucourt:
This also broke my existing pipelines which relied on configurable Credentials. We used to have a non existent 'None' credential by default as a placeholder.
When upgrading over [360.vf8027a_b_0c912](https://updates.jenkins.io/download/plugins/ssh-agent/360.vf8027a_b_0c912/ssh-agent.hpi), people relying on non existing Jenkins credentials should update their pipeline files with `ignoreMissing` like so:
```groovy
// Old
sshagent(credentials: [ExampleCred1, ExampleCred2]) {
    ...
}
// New
sshagent(credentials: [ExampleCred1, ExampleCred2], ignoreMissing: true) {
    ...
}
```

jglick:
(added a warning to https://github.com/jenkinsci/ssh-agent-plugin/releases/tag/360.vf8027a_b_0c912)

PR #152, opened 2024-07-29
Simplified plugin impl

jglick:
Cleaning up after #48.


pom.xml
```diff
@@ -57,7 +57,8 @@
   <properties>
     <changelist>999999-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
-    <jenkins.version>2.387.3</jenkins.version>
+    <jenkins.baseline>2.440</jenkins.baseline>
```
jglick, line 60: Just updating to a recommended baseline while we are here.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/LauncherProvider.java
```diff
@@ -1,20 +0,0 @@
...
-
-import edu.umd.cs.findbugs.annotations.NonNull;
-import hudson.Launcher;
-
-import java.io.IOException;
-
-/**
- * Provider of launcher objects. Needed in situations where the launcher object
- * might change in between usages and where it is not possible to propagate
- * the object in other ways.
```
jglick, line 11: We can just pass the `Launcher` to each call.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteAgent.java
```diff
@@ -1,60 +0,0 @@
...
- */
-
-package com.cloudbees.jenkins.plugins.sshagent;
-
-import hudson.model.TaskListener;
-
-import java.io.IOException;
-
-/**
- * Need an interface in order to export the object from the channel.
```
jglick, line 32: Not sure if this was actually true prior to #48, but it is certainly not true now—this interface was not being passed to `Channel.export`.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/RemoteAgentFactory.java
```diff
@@ -1,64 +0,0 @@
...
-package com.cloudbees.jenkins.plugins.sshagent;
-
-import edu.umd.cs.findbugs.annotations.CheckForNull;
-import hudson.ExtensionPoint;
-import hudson.FilePath;
-import hudson.Launcher;
-import hudson.model.TaskListener;
-
-/**
- * Extension point for ssh-agent providers.
```
jglick, line 34: When there is just a single impl for years, this was unnecessary complexity.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
```diff
@@ -296,95 +293,20 @@ public String getDisplayName() {
...
-        /**
-         * Construct the environment and initialize on the remote node.
-         *
-         * @param launcher          the launcher for the remote node.
-         * @param listener          the listener for reporting progress.
-         * @param sshUserPrivateKey the private key to add to the agent.
-         * @throws Throwable if things go wrong.
-         * @deprecated use {@link #SSHAgentEnvironment(hudson.Launcher, hudson.model.BuildListener, java.util.List)}
-         */
-        @Deprecated
```
jglick, line 318: No reason to keep these long-deprecated overloads around; who would call them?

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
```diff
@@ -296,95 +293,20 @@ public String getDisplayName() {
...
-                    StringWriter sw = new StringWriter();
-                    fault.getValue().printStackTrace(new PrintWriter(sw));
-                    for (String line : StringUtils.split(sw.toString(), "\n")) {
-                        listener.getLogger().println("[ssh-agent]     " + line);
-                    }
-                }
-                throw new RuntimeException("[ssh-agent] Could not find a suitable ssh-agent provider.");
-            }
-            this.agent = agent;
+            agent = new ExecRemoteAgent(launcher, listener);
```
jglick, line 309: When there is no list of factories to consult, the code is simpler and the error handling should be as well: either running `ssh-agent` works, or it does not.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
```diff
@@ -408,7 +330,7 @@ public void add(SSHUserPrivateKey key) throws IOException, InterruptedException
          */
         @Override
         public void buildEnvVars(Map<String, String> env) {
-            env.put("SSH_AUTH_SOCK", agent.getSocket());
+            env.putAll(agent.getEnv());
```
jglick, line 333: Probably only this one variable really _needs_ to be passed to nested steps, but `ssh-agent(1)` suggests you just `eval` the env, so why not?

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStep.java
```diff
@@ -6,26 +6,30 @@
...
+import java.util.Set;
+import org.jenkinsci.plugins.workflow.steps.Step;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
+import org.jenkinsci.plugins.workflow.steps.StepExecution;
 
-public class SSHAgentStep extends AbstractStepImpl implements Serializable {
-
-    private static final long serialVersionUID = 1L;
+public class SSHAgentStep extends Step {
```
jglick, line 32: Getting rid of long-deprecated Guice usage.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStep.java
```diff
@@ -86,7 +95,7 @@ public ListBoxModel doFillCredentialsItems(@AncestorInPath Item item) {
             }
             return new StandardUsernameListBoxModel()
                     .includeMatchingAs(
-                            item instanceof Queue.Task ? Tasks.getAuthenticationOf((Queue.Task)item) : ACL.SYSTEM,
+                            item instanceof Queue.Task ? Tasks.getAuthenticationOf2((Queue.Task)item) : ACL.SYSTEM2,
```
jglick, line 98: Spring Security

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -71,27 +41,20 @@ public boolean start() throws Exception {
...
-            purgeSockets();
-            initRemoteAgent();
-        } catch (IOException | InterruptedException x) {
-            listener.getLogger().println(Messages.SSHAgentBuildWrapper_CouldNotStartAgent());
-            x.printStackTrace(listener.getLogger());
+            stop();
+        } catch (Exception x) {
+            cause.addSuppressed(x);
         }
+        super.stop(cause);
```
jglick, line 49: Probably better, so we pass along interruptions and so on as is.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -71,27 +41,20 @@ public boolean start() throws Exception {
...
-        }
-        purgeSockets();
-    }
-
-    @Override
-    public void onResume() {
-        super.onResume();
         try {
-            purgeSockets();
-            initRemoteAgent();
```
jglick, line 86: This never made sense, and probably contributed to the instability of the plugin. Why would we stop and restart the SSH agent process, running on some agent computer, just because the controller restarted?

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -149,97 +115,17 @@ private void initRemoteAgent() throws IOException, InterruptedException {
...
-                listener.getLogger().println(Messages.SSHAgentBuildWrapper_Stopped());
-            }
-        } finally {
-            purgeSockets();
-        }
-    }
-
-    /**
-     * Purges all socket files created previously.
-     * Especially useful when Jenkins is restarted during the execution of this step.
```
jglick, line 211: Something else dating to #4 but of unclear purpose. We already call `ssh-agent -k`, which deletes the socket file.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java
```diff
@@ -149,97 +115,17 @@ private void initRemoteAgent() throws IOException, InterruptedException {
...
-     */
-    private void purgeSockets() {
-        Iterator<String> it = sockets.iterator();
-        while (it.hasNext()) {
-            File socket = new File(it.next());
-            if (socket.exists()) {
```
jglick, line 217: Furthermore, this was looking up the socket path on the _controller_ filesystem, where it would never exist. So this was dead code AFAICT.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
```diff
@@ -24,76 +24,58 @@
...
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
-
 /**
- * An implementation that uses native SSH agent installed on a system.
+ * Runs a native SSH agent installed on a system.
  */
-public class ExecRemoteAgent implements RemoteAgent {
+public final class ExecRemoteAgent implements Serializable {
```
jglick, line 43: OK to serialize now that this merely contains a `Map<String, String>`, and means that the `StepExecution` does not need to recreate it.

src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgentFactory.java
```diff
@@ -1,86 +0,0 @@
...
-        return "Exec ssh-agent (binary ssh-agent on a remote machine)";
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public boolean isSupported(Launcher launcher, final TaskListener listener) {
-        try {             
-            int status = launcher.launch().cmds("ssh-agent", "-k").quiet(true).start().joinWithTimeout(1, TimeUnit.MINUTES, listener);
```
jglick, line 60: Pointless overhead; we are just going to run it for real shortly afterwards.

julieheard (approved):

src/main/java/com/cloudbees/jenkins/plugins/sshagent/LauncherProvider.java
MikeyCarter, line 1: This file being replaced by something.  I've got failures all over the place with the removal of this.

jglick: Can you be more specific please? Do you have a Pipeline build which was in the middle of `sshagent` at the time of an upgrade? I do not think there is a functional test for that yet…

MikeyCarter: After the server is restarted, with the latest release from a few hours ago of ssh-agent, I have other plugins failing on start because they are looking for LauncherProvider class.   Notice in your latest merges/release this class has been removed.

So trying to determine if this LauncherProvider has been replaced with something so I can at the very least alter my own plugin to use the new setup.

jglick: > other plugins failing on start because they are looking for LauncherProvider

Which plugins? There are no outside references to this class that I am aware of.

> if this LauncherProvider has been replaced with something

No.

MikeyCarter: So yes there are some outside references.   I think I see how to modify my own code by looking at the changes.   If I figure out the other plugins I'll let you know, chances are if your unaware it's going to be something unregistered.

jglick: > I do not think there is a functional test for [a Pipeline build which was in the middle of `sshagent` at the time of an upgrade] yet

Tried to write one, but it looked hard since the plugin depends on a specific process running. Tested interactively and the step survived upgrade.

PR #163, opened 2024-11-19
Fixed end of substring in getAgentValue

aralex:
Fixed end of substring in `getAgentValue`. Otherwise the following error may occour:
```
[Pipeline] End of Pipeline
Also:   org.jenkinsci.plugins.workflow.actions.ErrorAction$ErrorId: f2e82224-14bd-4013-bd9e-f7f08a73c6c5
java.lang.StringIndexOutOfBoundsException: begin 13, end -1, length 0
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4601)
	at java.base/java.lang.String.substring(String.java:2704)
```

I found it in job on Windows agent.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
@@ -143,6 +143,7 @@ public final class ExecRemoteAgent implements Serializable {
     private String getAgentValue(String agentOutput, String envVar) {
         int pos = agentOutput.indexOf(envVar) + envVar.length() + 1; // +1 for '='
         int end = agentOutput.indexOf(';', pos);
+        if(end < 0) end = agentOutput.length();
         return agentOutput.substring(pos, end);
     }
     


jglick:
What is the text output that was misparsed on Windows? This looks like something that would benefit from test coverage.

aralex:
> What is the text output that was misparsed on Windows? This looks like something that would benefit from test coverage.

I don't know how to view the problem text. I thought to rebuild this java-file but I have no experience for it.

aralex:
This is the bigger log piece:
```[Pipeline] End of Pipeline
Also:   org.jenkinsci.plugins.workflow.actions.ErrorAction$ErrorId: f2e82224-14bd-4013-bd9e-f7f08a73c6c5
java.lang.StringIndexOutOfBoundsException: begin 13, end -1, length 0
	at java.base/java.lang.String.checkBoundsBeginEnd(String.java:4601)
	at java.base/java.lang.String.substring(String.java:2704)
	at com.cloudbees.jenkins.plugins.sshagent.exec.ExecRemoteAgent.getAgentValue(ExecRemoteAgent.java:146)
	at com.cloudbees.jenkins.plugins.sshagent.exec.ExecRemoteAgent.parseAgentEnv(ExecRemoteAgent.java:130)
	at com.cloudbees.jenkins.plugins.sshagent.exec.ExecRemoteAgent.<init>(ExecRemoteAgent.java:57)
	at com.cloudbees.jenkins.plugins.sshagent.SSHAgentStepExecution.initRemoteAgent(SSHAgentStepExecution.java:121)
	at com.cloudbees.jenkins.plugins.sshagent.SSHAgentStepExecution.start(SSHAgentStepExecution.java:35)
```
I would try to check the command manually. What command should I enter?

jglick:
https://github.com/jenkinsci/ssh-agent-plugin/blob/95d6ea23e3122e004eb8ae4a2345df6820814891/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java#L52-L57 so just try running `ssh-agent` on your agent machine and see what format the output is in.

aralex:
When I just run `ssh-agent` it prints nothing.

aralex:
I'm trying to work with Windows Server 2019. SSH agent works as a service there. And `ssh-agent` command does return nothing.

jglick:
> Don't the plugin use `PATH` environment variable to run `ssh-agent`?

https://github.com/jenkinsci/ssh-agent-plugin/blob/107d2461549a71714220542944f74d48b53ebd80/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepExecution.java#L102-L121 does not currently pay attention to Pipeline overrides with `withEnv('PATH+EXTRA=…') {…}`. It could be made to do so I suppose, just never came up. If you fix the path for `java -jar agent.jar` that would be honored.

Not sure what Microsoft’s version of the tool does, if it is not printing anything. Would be nice if the plugin automatically detected this tool and at least failed with a clear error.

HannesWell:
> This plugin does nothing all that special. If you run into any issues, just stop using it and run `ssh-agent` directly, or more commonly simply pass some option like `-i` to whatever command needs a private key.

That's relatable, but at the same time it's also not a satisfactory situation from a users perspective if this plugin doesn't work on Windows at all (as far as I understand it), while there is a possible solution. Furthermore it's currently not easy to understand what the issue actually is.
I have now created a proposal that worked at least for me locally:
- https://github.com/jenkinsci/ssh-agent-plugin/pull/169

PR #169, opened 2025-11-22
Support execution on Windows by using ssh-agent provided by git

HannesWell:
Since Windows-10 a ssh-agent is installed by default on Windows systems. But that agent works differently than the ssh-agent known from UNIX systems and therefore attempts to use this plugin fail. From the error message the cause of the failure is not easy to understand. And of course not being able to use this plugin on Windows agents is in general a unsatisfactory situation.

Fortunately the git (for-windows) installation by default contains a ssh-agent implementation that behaves like the known UNIX implementation and runs on Windows. And a git installation is often available.

This change extends the ssh-agent plugin to attempt to locate the described ssh-agent installation provided by git, when running on Windows.
If no executable is found, either through a git installation or by searching the system's PATH variable, the execution fails with a meaningful message.

If you don't want a strict requirement on `git-client` I could also make that optional and prepare the code to handle a potential absence at runtime.

This hopefully avoids confusion like in
- https://github.com/jenkinsci/ssh-agent-plugin/pull/163

### Testing done

I tested this locally with the changes applied on a Windows computer and it worked out of the box.
If you think there should be more test cases I'm happy to provide them, but I'd need some instructions and specific demands.
I see that you execute the build pipeline (including tests?) on Windows too, but from my understanding of the problem, this plugin effectively doesn't work on Windows at all (before this change). So either the latter assumption is wrong or the tests cover scenarios.

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,6 +90,11 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>ssh-credentials</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>git-client</artifactId>
+      <optional>true</optional>
+    </dependency>
     <!-- jenkins dependencies -->
     <!-- test dependencies -->
     <dependency>
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
@@ -26,20 +26,26 @@ package com.cloudbees.jenkins.plugins.sshagent.exec;
 
 import hudson.AbortException;
 import hudson.FilePath;
+import hudson.Functions;
 import hudson.Launcher;
 import hudson.Launcher.ProcStarter;
 import hudson.model.TaskListener;
+import hudson.remoting.VirtualChannel;
 import hudson.slaves.WorkspaceList;
+import jenkins.security.MasterToSlaveCallable;
+
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import java.util.regex.Pattern;
 
 /**
  * Runs a native SSH agent installed on a system.
@@ -60,6 +66,8 @@ public final class ExecRemoteAgent implements Serializable {
     /** The path of the ssh-agent executable. */
     private final String sshAgentBin;
 
+    private final boolean isWindowsAgent;
+
     /**
      * Launches a native {@code ssh-agent}.
      *
@@ -75,7 +83,10 @@ public final class ExecRemoteAgent implements Serializable {
         // A timeout of zero (or less) comes from job configurations persisted before the timeout
         // was configurable; treat it as the default rather than timing out immediately.
         this.timeoutMinutes = timeoutMinutes > 0 ? timeoutMinutes : DEFAULT_TIMEOUT_MINUTES;
-        FilePath sshAgentPath = toSSHAgentPath(executablePath, launcher);
+        this.isWindowsAgent = isWindowsAgent(launcher, listener);
+        FilePath sshAgentPath = executablePath == null && isWindowsAgent
+                ? searchSSHAgentExeForWindows(launcher, listener)
+                : toSSHAgentPath(executablePath, launcher);
         this.sshAgentBin = Optional.ofNullable(sshAgentPath).map(FilePath::getParent)
                 .map(p -> p.getRemote() + (launcher.isUnix() ? '/' : '\\')).orElse("");
 
@@ -93,6 +104,48 @@ public final class ExecRemoteAgent implements Serializable {
                     "Not an ssh-agent executable path (filename must be ssh-agent(.exe)): " + executable);
         }
         return path;
+    }
+
+    private FilePath searchSSHAgentExeForWindows(Launcher launcher, TaskListener listener)
+            throws IOException, InterruptedException {
+        try { // Git plugin is optional, handle a potential absence
+            // Search for the absolute path to the home directory of a git installation at
+            // the agent computer. Just git(.exe) is insufficient, as it only searches PATH.
+            Optional<String> defaultGitHome = Optional.ofNullable(hudson.plugins.git.GitTool.getDefaultInstallation())
+                    .map(hudson.plugins.git.GitTool::getHome)
+                    .filter(home -> !"git".equals(home) && !"git.exe".equals(home));
+            if (defaultGitHome.isPresent()) {
+                Optional<FilePath> sshAgentExe = extractGitSSHAgentExe(List.of(defaultGitHome.get()), launcher);
+                if (sshAgentExe.isPresent()) {
+                    return sshAgentExe.get();
+                }
+            }
+        } catch (NoClassDefFoundError e) { // git plugin is absent -> ignore
+        }
+        // Search a local git installation
+        String gitPaths = executeCommand(p -> p.cmds("where", "git").quiet(true), launcher, listener, true);
+        Optional<FilePath> sshAgentExe = extractGitSSHAgentExe(gitPaths.lines().filter(l -> !l.isEmpty())::iterator,
+                launcher);
+        return sshAgentExe.orElseThrow(() -> new IllegalStateException(
+                "Executing with default ssh-agent on Windows is not supported and an alternative implementation from a git installation is not available."));
+    }
+
+    private static final String GIT_EXE_PATH = "\\cmd\\git";
+    private static final String GIT_SSH_AGENT_PATH_WINDOWS = "usr\\bin\\ssh-agent.exe";
+
+    private static Optional<FilePath> extractGitSSHAgentExe(Iterable<String> gitHomeOrExePaths, Launcher launcher)
+            throws IOException, InterruptedException {
+        for (String path : gitHomeOrExePaths) {
+            Optional<FilePath> git = Optional.of(new FilePath(launcher.getChannel(), path));
+            if (path.endsWith(GIT_EXE_PATH + ".exe") || path.endsWith(GIT_EXE_PATH)) {
+                git = git.map(FilePath::getParent).map(FilePath::getParent);
+            }
+            Optional<FilePath> sshAgentExe = git.map(p -> p.child(GIT_SSH_AGENT_PATH_WINDOWS));
+            if (sshAgentExe.isPresent() && sshAgentExe.get().exists()) {
+                return sshAgentExe;
+            }
+        }
+        return Optional.empty();
     }
 
     private String executeCommand(Consumer<ProcStarter> processConfig, Launcher launcher, TaskListener listener,
@@ -209,7 +262,14 @@ public final class ExecRemoteAgent implements Serializable {
         // TODO better to just parse all env vars and keep them without picking out individual keys
 
         // get SSH_AUTH_SOCK
-        env.put(AuthSocketVar, getAgentValue(agentOutput, AuthSocketVar));
+        String socketPath = getAgentValue(agentOutput, AuthSocketVar);
+        if (isWindowsAgent) {
+            // Convert socket path (originally in linux-style) into Windows-style.
+            // The ssh-agent tools of a git installation can handle Windows (and Linux)
+            // style paths. But other Windows tools can only handle Windows-style.
+            socketPath = toWindowsPath(socketPath);
+        }
+        env.put(AuthSocketVar, socketPath);
         listener.getLogger().println(AuthSocketVar + "=" + env.get(AuthSocketVar));
 
         // get SSH_AGENT_PID
@@ -218,7 +278,17 @@ public final class ExecRemoteAgent implements Serializable {
         
         return env;
     }
-    
+
+    private static final Pattern WINDOWS_DRIVE_LETTER_IN_UNIX_PATH = Pattern.compile("^/[a-zA-Z]/");
+
+    private static String toWindowsPath(String socketPath) {
+        if (WINDOWS_DRIVE_LETTER_IN_UNIX_PATH.matcher(socketPath).find()) {
+            char driveLetter = Character.toUpperCase(socketPath.charAt(1));
+            socketPath = driveLetter + ":\\" + socketPath.substring(3);
+        }
+        return socketPath.replace('/', '\\');
+    }
+
     /**
      * Parses a value from ssh-agent output.
      *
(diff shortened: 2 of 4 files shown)


jglick:
Sorry, I am not maintaining this plugin save for critical fixes or routine org-wide maintenance. It works on Linux but was never intended for use on Windows, and is best avoided generally. https://github.com/jenkinsci/ssh-agent-plugin/pull/163#issuecomment-3427371309

HannesWell:
> Sorry, I am not maintaining this plugin save for critical fixes or routine org-wide maintenance. 

Too bad :/ Is there anybody else maintaining this too and interested in reviewing this?

> It works on Linux but was never intended for use on Windows, and is best avoided generally.

Understand, still I think it would be convenient if this would work on Windows too, especially in multi-OS pipelines where similar scripts run under multiple OS. Mainly because this plugin makes the use of SSH very convenient and simple.

> Another alternative or complementary solution could be adding a configuration option with the ssh-agent executable.
> The default would be just `ssh-agent`, but one could also specify an absolute path to any other exe.

Or would you be willing to accept at least a change to just add the proposed `executable` parameter (or however you'd like to name it)? This way this plugin could at least be made to work on Windows without modifying the entire system PATH of the executing machine. And the change should be small and easy to review.

jglick:
> Is there anybody else maintaining this

I do not think so. I will mark it for adoption. (See #153.)

HannesWell:
> > Is there anybody else maintaining this
> 
> I do not think so. I will mark it for adoption. 

Too bad :/


> See #153.

Thanks for the pointer, I looked into this but I'm struggling to use the approach using `eval $(ssh-agent -s)` for a `sshUserPrivateKey` credential protected by a password.
Do you have any hints how to set up the ssh-agent securely in that case?

jglick:
On Windows? No. For most purposes in CI systems you would not need to use `ssh-agent` at all, just directly pass a private key to some command.

jimklimov:
> For most purposes in CI systems you would not need to use ssh-agent at all, just directly pass a private key to some command.

To be fair, my dayjob pipelines do a fair amount of work on SUT machines over SSH (without a jenkins agent running on them) so passing stacks of credentials and an `ssh-agent` environment (and then a `sh` step calling `ssh` with some script on stdin) is a common occurrence. Use-cases for CI systems vary widely. The plugin takes care of the key files not being in the same directory tree as tested code, so that is beneficial too (e.g. for tests that `git status` after a build does not show up surprising not-ignored not-tracked files).

@HannesWell : now that the plugin is officially up for adoption, you have an easy path to becoming actually a maintainer, and merging your fix as a first step in this career :D
See https://www.jenkins.io/doc/developer/plugin-governance/adopt-a-plugin/ for details.

ogulcanaydogan:
Thanks for this, and sorry it sat so long. Windows support is something the plugin should have, and reusing the ssh-agent from Git for Windows is a sensible way to get UNIX-like behaviour.

`ExecRemoteAgent` has changed quite a bit since November (clearer failure messages, best-effort teardown, and a configurable timeout that shipped in 405) and the constructor was restructured, so this no longer merges. Could you rebase onto master? Glad to help if the conflicts get messy.

Two things while you're in there:
- The constructor picks the Git ssh-agent on Windows, but `addIdentity` still runs `ssh-add` and `stop` runs `ssh-agent -k`. Do those find the Git binaries too once Git is on PATH, or do they need the same lookup?
- On the strict `git-client` dependency, I'd lean towards making it optional like you offered, so we don't force it on non-Git setups. Fine by you?

A small Windows test would be great too, since our CI runs on Windows and it'd keep this from regressing.

HannesWell:
> now that the plugin is officially up for adoption, you have an easy path to becoming actually a maintainer, and merging your fix as a first step in this career :D

While I'd really like to have this change included, unfortunately I have too many other things to do and this entire plugin is also not my main area of expertize. And just becoming maintainer for one change wouldn't be sustainable and wouldn't feel fair for the community.

That being said, I'm glad took over Ogulcan and considers this change.

> Thanks for this, and sorry it sat so long. Windows support is something the plugin should have, and reusing the ssh-agent from Git for Windows is a sensible way to get UNIX-like behaviour.

I'm glad you like this change. I'm working on the rebase and will also make the git dependency optional and will handle it's absence. I'll also check you requests and will try to add a test-case.

One more question:
What do you think about making the _executable_ configurable too, as proposed in https://github.com/jenkinsci/ssh-agent-plugin/pull/169#issuecomment-3566853237?
Not only would this give users even more flexibility to e.g. use a completely different exe, it could also be used as way to suppress the automatic search on windows by just setting the exe, which could even just be the `ssh-agent` default.

ogulcanaydogan:
Thanks, this is really generous of you, especially taking the rebase, making the git dependency optional and adding a test. No rush on any of it.

On the configurable executable: yes, I really like that idea. It is a clean escape hatch on top of the Windows auto-detection, someone can point it at a different binary, or just set it to `ssh-agent` to skip the search entirely.

I would keep it optional with today's behaviour as the default (plain `ssh-agent` on Unix, auto-detect on Windows) so it stays additive. Happy to have it in this PR if it does not grow the diff too much, otherwise a follow-up works just as well.

And no worries at all about not wanting to maintain it, that is completely fair. Thanks again for getting this going and for coming back to it.

HannesWell:
> On the configurable executable: yes, I really like that idea. It is a clean escape hatch on top of the Windows auto-detection, someone can point it at a different binary, or just set it to `ssh-agent` to skip the search entirely.
> 
> I would keep it optional with today's behaviour as the default (plain `ssh-agent` on Unix, auto-detect on Windows) so it stays additive. Happy to have it in this PR if it does not grow the diff too much, otherwise a follow-up works just as well.

I think it makes sense to separate the changes to make it easier to review and test and to follow for anybody (maybe later).
If started to create the first PRs for prerequisite changes.

While working on this I encountered some fundamental issues, that I've to investigate first.
I cannot tell why it worked before or if the assumed working state was not truly working.
But I'll keep you posted.

ogulcanaydogan:
Thanks for pushing the executable property, that is the escape hatch we discussed. The PR is conflicting with master now, so CI cannot build the merge, most likely from the recent merges and the release moving master forward. Could you rebase onto master? I will give it a full review once it is green.

HannesWell:
Thanks. Just rebased it.

In my local testing the current state looked good and worked well.
I was able to clone a repository via SSH without being asked for a password (in the build console), which happens without the `sshAgent` step.
Unfortunately I couldn't conduct further local testing.
But since all tests now run on Windows too and non of the previous disablements on Windows is still active, I hope that these tests cover all relevant use-cases.

The latest change should make spotbugs happy (although I'm convinced that the flagged null access cannot happen).
 If there are test-failures on Windows, I hope that I can see sufficient information in the GH UI.
If not, could you copy over the detailed test failure reports (if any) over from Jenkins and post it here?
Edit: The latest test results are clear to me, even from just the GH UI.

ogulcanaydogan:
Thanks for the rebase, and for turning the tests on for Windows, that is the right way to prove this out.

I had a look at the run so you don't have to chase it alone. windows-21 is at 12 failed / 30 passed, and they line up with the workflow tests where the assumeFalse(isWindows) guards came off. A couple of things I noticed that might save you some time:

- sshAgentDocker (SECURITY-704) spins up a Docker container, so it can't really pass on a Windows agent. That one probably wants to keep its Windows skip rather than count as a real failure.
- The rest (sshAgentAvailable, sshAgentAvailableAfterRestart, testTrackingOfCredential and friends) exercise the actual ssh-add, askpass and socket-path handling on Windows, which is exactly your new code, so those are the interesting ones to look at.

I don't have a Windows box in front of me to reproduce, but I'm glad to pair on it. If you point me at one or two of the failing assertions I'll dig into the code side with you. No rush at all, this is a genuinely useful addition and I'd like to land it properly.

ogulcanaydogan:
I set up a real Windows 11 box and reproduced the failures so you had concrete data to work from. All six failing SSHAgentStepWorkflowTest cases share one root cause, and it is not the git-agent detection: that part works perfectly, `"C:\Program Files\Git\usr\bin\ssh-agent"` resolves fine.

They all fail in the ssh-add step, and only because the Jenkins test harness deliberately puts a space in the workspace path (`...\target\tmp\j h1666...\...`). With that space, ssh-add exits 1 with:

```
'C:\Users\...\target\tmp\j' is not recognized as an internal or external command
```

so a space-containing path is reaching cmd unquoted. The likely culprit is the Windows askpass `.bat` (or the ssh-add invocation) not surviving spaces in the path. On a workspace path without a space it would pass, which is why local runs looked clean.

Full disclosure: the Windows box is my mother-in-law's laptop, which I borrowed for the afternoon, so she gets partial credit for the repro. Happy to test any fix you push on the same machine, just ping me.

HannesWell:
Thanks for the analysis.

> * sshAgentDocker (SECURITY-704) spins up a Docker container, so it can't really pass on a Windows agent. That one probably wants to keep its Windows skip rather than count as a real failure.


Sounds reasonable. But I'll leave the disablement to the very end to make sure that nothing is ignored prematurely.

> They all fail in the ssh-add step, and only because the Jenkins test harness deliberately puts a space in the workspace path (`...\target\tmp\j h1666...\...`). With that space, ssh-add exits 1 with:
> 
> ```
> 'C:\Users\...\target\tmp\j' is not recognized as an internal or external command
> ```
> 
> so a space-containing path is reaching cmd unquoted. The likely culprit is the Windows askpass `.bat` (or the ssh-add invocation) not surviving spaces in the path. On a workspace path without a space it would pass, which is why local runs looked clean.


Yes, I was able to just reproduce that case and also already tried out a few things.
I can already sort out the self-deletion command since, at least the same failure happens if the ask-pass script doesn't have the self deletion built-in (maybe it becomes a problem later).

After some more testing I can confirm that the path read from the `SSH_ASKPASS` environment variable is not properly escaped when being executed as command, probably within the ssh-add command.
If I set it to a fixed local path at my computer (without spaces) it works well.
So I'll have to check how to make the ssh-add command to properly escape the `SSH_ASKPASS` path.

But I'll continue with that tomorrow.


> Full disclosure: the Windows box is my mother-in-law's laptop, which I borrowed for the afternoon, so she gets partial credit for the repro. Happy to test any fix you push on the same machine, just ping me.

It's all family business :D Thanks to her too.

HannesWell:
> After some more testing I can confirm that the path read from the `SSH_ASKPASS` environment variable is not properly escaped when being executed as command, probably within the ssh-add command.
> So I'll have to check how to make the ssh-add command to properly escape the `SSH_ASKPASS` path.

After more in depth investigation and reproducing the situation with git's ssh-agent directly I can confirm that it's general issue and the `ssh-add` command does not handle the execution of the askpass command properly if it contains spaces. This has been reported to git-for-windows
- https://github.com/git-for-windows/git/issues/6366

Maybe it's even an upstream issue in the OpenSSH ssh-add implementation.

In order to work around this limitation I adjusted the creation of the askpass script to create the file not in the temporary directory associated with the workspace, but within the temp folder of the agent's JVM (i.e. it's `java.io.tmpdir`), hoping that that path doesn't contain spaces.

Furthermore I noticed that some file-path parsing and file-system operations that are supposed to happen at/for the agent computer and where actually done for the controller, so both where not properly distinguished.
I've fixed that for already existing code in
- https://github.com/jenkinsci/ssh-agent-plugin/pull/304

That change is currently also included in this PR.

ogulcanaydogan:
This is a big one, thank you for sticking with it, real Windows ssh-agent support closes out a bunch of long-standing pain (#218, #192, #163 and friends). The approach of locating git-for-windows' bundled ssh-agent when no executable is configured makes a lot of sense given how commonly git is already present on Windows agents, and using MasterToSlaveCallable to detect the OS on the actual agent rather than the controller is exactly right.

Really glad to see the previously-skipped Windows tests in SSHAgentStepWorkflowTest passing on CI now, including sshAgentAvailableAfterRestart, that one's been a real pain (see #298, the socket path regex was one part of it, the SSH_ASKPASS path handling under a workspace path with a space was the other, so seeing it green with the askpass.bat rewrite is a good sign that's actually resolved).

Two things before I merge this:
1. Same as #304, this also touches toSSHAgentPath, and #306 (open now, CI running) changes that same null-check for a different bug. Small rebase needed once #306 lands, happy to do it myself if you'd rather not deal with it.
2. Given #304 is essentially the FilePath portion of this PR split out on its own, do you want #304 merged first and this rebased on top (probably shrinks this diff a bit), or should I just take this one as-is and let #304 get closed as superseded? Either way works for me, your call since you're the one who split it.

ogulcanaydogan:
Same as I mentioned on #304, I rebased this one too and pushed it to my own fork rather than touch your branch: https://github.com/ogulcanaydogan/ssh-agent-plugin/tree/rebase-169-onto-master

Same one-line conflict in toSSHAgentPath, resolved the same way, everything else applied cleanly. Full suite green (50/50) including the previously-Windows-skipped tests. No pressure to use it, entirely your call whether you want to cherry-pick from there, rebase yourself, or ignore it, just didn't want the rebase to sit as extra work on your side since I'm the one who caused it.

HannesWell:
> This is a big one, thank you for sticking with it, real Windows ssh-agent support closes out a bunch of long-standing pain (#218, #192, #163 and friends). The approach of locating git-for-windows' bundled ssh-agent when no executable is configured makes a lot of sense given how commonly git is already present on Windows agents

I'm glad to and thanks for having, reviewing and supporting this.

Just rebased this on top of master, with https://github.com/jenkinsci/ssh-agent-plugin/pull/304 again included.
But as you suggested in your second point, I'd also prefer to have https://github.com/jenkinsci/ssh-agent-plugin/pull/304 submitted in advance to keep this PR's eventual diff smaller.

I can rebase again, if needed.

ogulcanaydogan:
#304 and #305 are both merged now. Whenever you get a chance, go ahead and rebase this on top so the diff shrinks down to just the Windows-specific parts, that'll make it much easier to review properly. No rush, take your time.

ogulcanaydogan:
Actually, went ahead and rebased this myself onto the latest master (now that #304 and #305 are both merged) and pushed directly to this branch, since maintainer edits are enabled. CI is running now. No need for you to do anything here, just didn't want you to have to rebase a third time over something this mechanical.

HannesWell:
Thank you very much for your help and reviews.
I'm happy that this is now integrated and I'm looking forward to try and use this in the real world.

@ogulcanaydogan Could please you make sure that a new release with this change is published? Thanks in advance.

ogulcanaydogan:
Done, 424 is out with this in it. Glad this is landing, and thanks again for pushing it through, the Windows support has been a long-standing gap. Let us know how it goes in real use.

HannesWell:
> Let us know how it goes in real use.

Sorry it took a while, the IT needed some time to install the update first and then I was busy with other task.
But now, after the follow-up changes done in
- https://github.com/jenkinsci/ssh-agent-plugin/pull/316
- https://github.com/jenkinsci/ssh-agent-plugin/pull/314
- https://github.com/jenkinsci/ssh-agent-plugin/pull/312
- https://github.com/jenkinsci/ssh-agent-plugin/pull/310
- https://github.com/jenkinsci/ssh-agent-plugin/pull/307

...  were applied, everything works fine and the `sshagent` step just works out of the box, even on Windows.

Thanks for fixing the reported bugs relating to this and reviewing this and all the other changes.

PR #282, opened 2026-07-09
fix: include stderr and exit code when ssh-agent fails

ogulcanaydogan:
### Problem
When `ssh-agent` exits with a non-zero status, `ExecRemoteAgent`'s constructor only captured stdout, so the resulting `AbortException` was usually `Failed to run ssh-agent: ` with nothing after the colon. `ssh-agent` normally writes the actual failure to stderr, which was discarded, leaving users with no way to tell what went wrong. This is issue #278 (JENKINS-74889).

### Fix
Capture stderr alongside stdout and always include the process exit code in the message, falling back to stdout when stderr is empty. The message is assembled by a small package-private `static` helper (`describeFailure`) so the formatting is covered by a unit test.

### Testing
Added `ExecRemoteAgentFailureMessageTest` asserting the message includes stderr when present, falls back to stdout, and still reports the exit code when there is no output at all (the previously useless case). `mvn test -Dtest=ExecRemoteAgentFailureMessageTest` passes. The success path is unchanged.

Fixes #278


jglick:
This plugin is using CD. That means you must apply a single categorization label to a PR before merging it.

(https://github.com/jenkins-infra/interesting-category-action/issues/1)

PR #284, opened 2026-07-17
fix: make ssh-agent teardown best-effort

ogulcanaydogan:
Stopping the agent runs `ssh-agent -k` during teardown. If the agent process was already gone (for example killed during a long build), the non-zero exit threw an `AbortException` that failed an otherwise successful build (JENKINS-43716).

This logs the failure instead of throwing, so teardown is best-effort. The message still includes the exit code and stderr, so a genuine problem is still visible in the build log.

### Testing

Added `teardownDoesNotFailBuildWhenAgentAlreadyStopped` to `SSHAgentStepWorkflowTest`: it stops the agent from inside the `sshagent` block, so the automatic teardown finds it already gone, and asserts the build still succeeds and the failure is logged. Without the change the build fails on teardown. `mvn test -Dtest=SSHAgentStepWorkflowTest#teardownDoesNotFailBuildWhenAgentAlreadyStopped,ExecRemoteAgentFailureMessageTest` passes.

Fixes #230

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
+++ b/src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
@@ -87,6 +87,40 @@ public class SSHAgentStepWorkflowTest extends SSHAgentBase {
     }
 
     /**
+     * Verifies that a failed {@code ssh-agent -k} during teardown does not fail an otherwise
+     * successful build. The build stops the agent from inside the block, so the automatic teardown
+     * finds it already gone; with best-effort teardown the build still succeeds and the failure is
+     * only logged (JENKINS-43716).
+     */
+    @Test
+    public void teardownDoesNotFailBuildWhenAgentAlreadyStopped() throws Exception {
+        assumeFalse(Functions.isWindows());
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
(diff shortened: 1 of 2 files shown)


src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
```diff
@@ -86,6 +86,40 @@ public void evaluate() throws Throwable {
...
+                WorkflowJob job = story.j.jenkins.createProject(WorkflowJob.class, "teardownBestEffort");
+                job.setDefinition(new CpsFlowDefinition(""
+                        + "node('" + story.j.createSlave().getNodeName() + "') {\n"
+                        + "  sshagent (credentials: ['" + CREDENTIAL_ID + "']) {\n"
+                        + "    sh 'ssh-agent -k'\n"
+                        + "  }\n"
+                        + "}\n", true)
+                );
+                WorkflowRun run = story.j.assertBuildStatusSuccess(job.scheduleBuild2(0));
+                story.j.assertLogContains("Failed to run ssh-agent -k", run);
```
jglick, line 118: #290

jglick: @ogulcanaydogan could you disable this test please, unless there is some straightforward way to detect the sort of environment in which it can pass?

PR #292, opened 2026-07-27
Add optional Git host key verification via git-client

ogulcanaydogan:
Adds an opt-in `hostKeyVerification` flag to the `sshagent` step (#265). When enabled, the block exposes `GIT_SSH_COMMAND` built from the git-client plugin's globally configured host key verification strategy, so Git operations verify the remote host key without manually injecting `~/.ssh/known_hosts`.

```groovy
sshagent(credentials: ['my-key'], hostKeyVerification: true) {
    sh 'git fetch ...'
}
```

### How it works

- Reads the global strategy: `GitHostKeyVerificationConfiguration.get().getSshHostKeyVerificationStrategy().getVerifier()`.
- On the agent (via a `FileCallable`), creates a temporary known_hosts and calls `forCliGit(listener).getVerifyHostKeyOption(knownHosts)`, mirroring what git-client's `CliGitAPIImpl` does.
- Exposes `GIT_SSH_COMMAND=ssh <options>` for the block and deletes the temporary known_hosts on teardown.
- Disabled by default, so existing pipelines are unaffected. Configured from the pipeline script (no UI change yet).

### Draft: needs review and validation

I am opening this as a draft because the host key behaviour needs validation I could not do locally (no real ssh server plus host key setup). I would appreciate review on:

- Whether the agent-side known_hosts handling is correct for every strategy (KnownHostsFile / AcceptFirstConnection / ManuallyProvidedKey / None), especially when the configured known_hosts lives on the controller rather than the agent.
- `GIT_SSH_COMMAND` collision: this currently overrides an existing value inside the block; should it respect or append a user-set one instead?
- Whether verification should be automatic (always use the global strategy) rather than opt-in.

The step config is unit tested; the host key wiring relies on CI integration tests and review.

Refs #265

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStep.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStep.java
@@ -64,6 +64,16 @@ public class SSHAgentStep extends Step {
     private String executable;
 
     /**
+     * If enabled, {@code GIT_SSH_COMMAND} is exposed inside the block so that Git operations verify
+     * the remote host key using the git-client plugin's globally configured verification strategy.
+     * The verifier's known_hosts is created on the agent where the build runs, so a
+     * {@code KnownHostsFile} strategy pointing at a path on the controller is not visible to the agent.
+     * A {@code GIT_SSH_COMMAND} the user already set is left untouched.
+     * Optional; disabled by default so existing pipelines are unaffected.
+     */
+    private boolean hostKeyVerification;
+
+    /**
      * Default parameterized constructor.
      *
      * @param credentials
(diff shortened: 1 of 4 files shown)


ogulcanaydogan:
Marking this ready. Rebased onto master and resolved the three open questions from the description:

- known_hosts handling mirrors git-client's CliGitAPIImpl, with the verify options computed on the agent. The one caveat, a KnownHostsFile strategy pointing at a controller path, is now documented on the field.
- GIT_SSH_COMMAND collision: it no longer replaces a value the user already set. It logs that host key verification was skipped and leaves the existing value in place.
- Kept it opt-in rather than automatic, to stay backward compatible.

Added two workflow tests: one checks that a pre-set GIT_SSH_COMMAND is preserved, the other that enabling it exposes a GIT_SSH_COMMAND for the block. I verified both on a real Linux box, and CI is green on the linux-25 and windows-21 lines. The tests assert the plugin's own behaviour rather than git-client's exact options, to avoid coupling to its version

PR #294, opened 2026-07-30
Make new timeout property usable in free-style jobs and add it to doc

HannesWell:
Add a timeout field to the SSHAgentBuildWrapper to make it usable in freestyle jobs, too.

Follow-up on
- https://github.com/jenkinsci/ssh-agent-plugin/pull/286

Personally I'm not using freestyle jobs, but for completes and feature parity, I think this would be good.

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
@@ -87,6 +87,13 @@ public class SSHAgentBuildWrapper extends BuildWrapper {
     private final boolean ignoreMissing;
 
     /**
+     * The timeout (in minutes) for ssh agent commands.
+     *
+     * @since FIXME
+     */
+    private final int timeout;
+
+    /**
      * Constructs a new instance.
      *
      * @param user the {@link SSHUserPrivateKey#getId()} of the credentials to use.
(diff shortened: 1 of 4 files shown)


HannesWell:
@ogulcanaydogan could you please review this?

ogulcanaydogan (commented):
Thanks for picking this up, having the timeout on freestyle jobs too makes sense.

One thing before it actually works: the `@DataBoundConstructor` drops the timeout. It calls `this(CredentialHolder.toIdList(credentialHolders), ignoreMissing)`, which routes to the deprecated `(List<String>, boolean)` overload, and that one fills in `DEFAULT_TIMEOUT_MINUTES`. So whatever the user types in the field never reaches `this.timeout`, it stays at 1. Passing it through fixes it:

```java
this(CredentialHolder.toIdList(credentialHolders), ignoreMissing, timeout);
```

Smaller note: the new `timeout` field/getter is `@since 405`, but 405 already shipped on Jul 22 without it, so this one is the next release (`@since FIXME` like #286 did). The FIXME to 405 changes on the ExecRemoteAgent side are right, that code did land in 405.

HannesWell:
> One thing before it actually works: the `@DataBoundConstructor` drops the timeout.

Oh yes, I missed that. Thanks for pointing out. Fixed it (as suggested).

> Smaller note: the new `timeout` field/getter is `@since 405`, but 405 already shipped on Jul 22 without it, so this one is the next release (`@since FIXME` like #286 did). The FIXME to 405 changes on the ExecRemoteAgent side are right, that code did land in 405.

Yes. I wanted to be a bit 'lazy' and avoid subsequent changes by using the version where the timeout feature generally became available. But changed it as you suggested.
And added javadoc for the new parameters in the constructors.

ogulcanaydogan (approved):
Nice, thanks for the quick fix. The DataBoundConstructor threads the timeout through now, and the `@since` split reads right (FIXME for the freestyle field, 405 for the ExecRemoteAgent bits that already shipped). Good call extending the timeout to freestyle jobs too, that was a real gap. Approving, thanks!

HannesWell:
Great, thanks for your reviews.

Can this and my other ready/approved PR thru be submitted? 
Thanks in advance.

ogulcanaydogan:
Happy to land all three! Since they overlap in `ExecRemoteAgent`, I will merge them one at a time and each following one will need a quick rebase. Starting with #295 (the refactor). Once it is in, could you rebase #294 on top? Then after #294 lands, #296 becomes just the executable delta. That keeps your three as distinct changes. Thanks for all the work on these.

HannesWell:
> Once it is in, could you rebase #294 on top? Then after #294 lands, #296 becomes just the executable delta.

Sounds good. Just rebased this one. I had a rebase butten in the GH UI.
The last of the three seems to need a manual rebase. Will do that later this evening.

PR #295, opened 2026-07-30
Unify sub-process command execution in ExecRemoteAgent

HannesWell:
Unify sub-process command execution in `ExecRemoteAgent`.
This avoids duplicated logic and simplifies subsequent changes.

### Testing done

Executed the new code in a local Jenkins instance.

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
@@ -27,6 +27,7 @@ package com.cloudbees.jenkins.plugins.sshagent.exec;
 import hudson.AbortException;
 import hudson.FilePath;
 import hudson.Launcher;
+import hudson.Launcher.ProcStarter;
 import hudson.model.TaskListener;
 import hudson.slaves.WorkspaceList;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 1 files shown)


HannesWell:
@ogulcanaydogan could you please review this?

ogulcanaydogan (commented):
Nice cleanup, the three call sites had drifted and this brings them back together.

One behavior change worth a look: on an `ssh-add` failure the message now comes from `String.join(" ", starter.cmds())`, so it includes the full `ssh-add <keyfile>` command line. The old path passed a bare `"ssh-add"` on purpose, since a few lines up we log "Running ssh-add (command line suppressed)" to keep that path out of the log. It is the temp file path, not the key, so minor, but it does undo that suppression on the error path. Rest looks good.

HannesWell:
> One behavior change worth a look: on an `ssh-add` failure the message now comes from `String.join(" ", starter.cmds())`, so it includes the full `ssh-add <keyfile>` command line. The old path passed a bare `"ssh-add"` on purpose, since a few lines up we log "Running ssh-add (command line suppressed)" to keep that path out of the log. It is the temp file path, not the key, so minor, but it does undo that suppression on the error path. Rest looks good.

That's correct. The main motivation was to have the full command available in case the shutdown fails. On shutdown the command is `ssh-agent -k`, but if only the first command is added it shows just  `ssh-agent` and would be equal to the startup case.
And since it's only included in the error case IMO it's fine, respectively in case of errors I think it's even better to be detailed (as long as one doesn't leak credentials, which isn't happening).

So I'd be in favor of keeping it as it is, but if you still prefer to include only the command/first value in the error message, I can change it.

ogulcanaydogan (approved):
Ah, good point about `ssh-agent -k` vs the startup case, I had not considered the shutdown side. You are right that on the error path the extra detail is worth more than the tidiness, and since it is just the temp file path there is no credential exposure. Happy to keep it as is. Thanks for the clean refactor, approving!

PR #296, opened 2026-07-31
Introduce executable property to use alternative executables

HannesWell:
Allows users to specify the `ssh-agent` executable used when launching and setting up the ssh-agent.
When calling `ssh-add`, that executable is expected to be co-located to the specified `ssh-agent` executable.

### Testing done

Conducted only local testing at the moment.
Introducing a test case could be difficult because probably only one ssh-agent is available on CI test machines.
But it just could be tested if using the only agent explicitly works.

Currently this is a draft, because it's based on and includes the changes of
- https://github.com/jenkinsci/ssh-agent-plugin/pull/294
- https://github.com/jenkinsci/ssh-agent-plugin/pull/295

once these PRs are merged, this can be rebased and would then be ready.

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
@@ -94,6 +94,13 @@ public class SSHAgentBuildWrapper extends BuildWrapper {
     private final int timeout;
 
     /**
+     * The path to the {@code ssh-agent} executable.
+     *
+     * @since FIXME
+     */
+    private final String executable;
+
+    /**
      * Constructs a new instance.
      *
      * @param user the {@link SSHUserPrivateKey#getId()} of the credentials to use.
@@ -107,13 +114,13 @@ public class SSHAgentBuildWrapper extends BuildWrapper {
 
     @SuppressWarnings("unused") // used via stapler
     public SSHAgentBuildWrapper(CredentialHolder[] credentialHolders, boolean ignoreMissing) {
-        this(credentialHolders, ignoreMissing, ExecRemoteAgent.DEFAULT_TIMEOUT_MINUTES);
+        this(credentialHolders, ignoreMissing, ExecRemoteAgent.DEFAULT_TIMEOUT_MINUTES, null);
     }
 
     @Deprecated
     @SuppressWarnings("unused") // used via stapler
     public SSHAgentBuildWrapper(List<String> credentialIds, boolean ignoreMissing) {
-        this(credentialIds, ignoreMissing, ExecRemoteAgent.DEFAULT_TIMEOUT_MINUTES);
+        this(credentialIds, ignoreMissing, ExecRemoteAgent.DEFAULT_TIMEOUT_MINUTES, null);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


ogulcanaydogan (commented):
This stacks on #294 and #295, so a couple of notes carry over, plus one specific to the executable handling.

The executable prefix breaks when the given path has no parent directory. `toSSHAgentPath` accepts a bare filename like `ssh-agent`, and then:

```java
this.sshAgentBin = sshAgentPath == null ? "" : sshAgentPath.getParent() + File.separator;
```

`Path.of("ssh-agent").getParent()` is null, so `sshAgentBin` becomes the literal string `"null/"` and the commands turn into `null/ssh-agent` and `null/ssh-add`. Guarding the null parent covers it:

```java
Path parent = sshAgentPath == null ? null : sshAgentPath.getParent();
this.sshAgentBin = parent == null ? "" : parent + File.separator;
```

Same as on #294: the `@DataBoundConstructor` drops both new fields. It calls `this(CredentialHolder.toIdList(credentialHolders), ignoreMissing)`, which routes to the deprecated 2-arg overload and fills in `DEFAULT_TIMEOUT_MINUTES, null`, so neither the freestyle timeout nor the executable reach the instance. It needs `this(CredentialHolder.toIdList(credentialHolders), ignoreMissing, timeout, executable)`.

Minor: the new `timeout` and `executable` fields are `@since 405`, but 405 already shipped without them, so that would be the next release (or FIXME like #286 used).

HannesWell:
Thanks for the review.
I believe I've addressed all your remarks.

> `Path.of("ssh-agent").getParent()` is null, so `sshAgentBin` becomes the literal string `"null/"` and the commands turn into `null/ssh-agent` and `null/ssh-add`. Guarding the null parent covers it:
> ```
> Path parent = sshAgentPath == null ? null : sshAgentPath.getParent();
> this.sshAgentBin = parent == null ? "" : parent + File.separator;
> ```

Indeed that's a major bug. Thanks for spotting.

I've addressed this with the one-liner:
```
this.sshAgentBin = Optional.ofNullable(sshAgentPath).map(p -> p.getParent() + File.separator).orElse("");
```
But if you prefer your suggestion, I can also change it to that.

ogulcanaydogan (commented):
Thanks, the DataBoundConstructor threading and the `@since` fixes look good now. One thing still on the executable path: the `Optional` handles a null executable, but a bare filename like `ssh-agent` (no directory) still trips it, since `Path.of("ssh-agent").getParent()` is null and `null + File.separator` gives the literal `"null/"`. Chaining getParent through the Optional covers it:

```java
this.sshAgentBin = Optional.ofNullable(sshAgentPath).map(Path::getParent).map(p -> p + File.separator).orElse("");
```

That way a null parent falls through to the empty prefix. Everything else looks great.

ogulcanaydogan:
Thanks! One catch though: the one-liner still trips on the bare filename. `Optional.ofNullable(sshAgentPath)` guards `sshAgentPath` being null, but for something like `ssh-agent` with no directory, sshAgentPath is non-null and `p.getParent()` returns null, so `null + File.separator` is still the literal `"null/"`. The chaining version handles it because `.map(Path::getParent)` turns a null parent into an empty Optional that falls through to `orElse("")`:

```java
this.sshAgentBin = Optional.ofNullable(sshAgentPath).map(Path::getParent).map(p -> p + File.separator).orElse("");
```

Mind switching to that one? Then it is fully covered.

HannesWell:
> One catch though: the one-liner still trips on the bare filename. `Optional.ofNullable(sshAgentPath)` guards `sshAgentPath` being null, but for something like `ssh-agent` with no directory, sshAgentPath is non-null and `p.getParent()` returns null, so `null + File.separator` is still the literal `"null/"`. The chaining version handles it because `.map(Path::getParent)` turns a null parent into an empty Optional that falls through to `orElse("")`:

Yes, of course. Seems like doing changes 'quickly' late at night makes one doing little stupid mistakes. 🤦🏽‍♂️
I believe I used the Optional approach with that chaining in mind, but somehow forgot it along the way...

Thanks again for paying attention to these (important) details.

ogulcanaydogan (approved):
That does it, thanks for bearing with the back and forth! Late-night one-liners get all of us. The chaining covers the bare filename now, and the DataBoundConstructor and `@since` changes look good too. Approving.

HannesWell:
This is now ready for submission as we planned in https://github.com/jenkinsci/ssh-agent-plugin/pull/294#issuecomment-5208857917.

Thanks for your reviews!

ogulcanaydogan:
All three are in now, thanks for the great work on these! The refactor and freestyle timeout shipped in 414, and the executable property will go out in the next release.

PR #298, opened 2026-08-03
test: match ssh-agent socket path independent of platform layout

### LINKED ISSUE #289: Current bug in tests that was discovered during the weekly BOM release for July 24, 2026
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
OS: macOS Tahoe 26.5.2 (25F84)
Maven: Apache Maven 3.9.16 (2bdd9fddda4b155ebf8000e807eb73fd829a51d5)
Jenkins: 2.574
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

macOS Tahoe 26.5.2 (25F84)

### Reproduction steps

Download the default `master` branch and then run `mvn clean verify`. 

### Expected Results

All tests pass

### Actual Results

[ERROR] Failures: 
[ERROR]   SSHAgentStepWorkflowTest.lambda$sshAgentAvailableAfterRestart$1:213 [] ==> expected: <2> but was: <0>
[INFO] 
[ERROR] Tests run: 39, Failures: 1, Errors: 0, Skipped: 1

### Anything else?

_No response_

### Are you interested in contributing a fix?

Maybe

ogulcanaydogan:
`SSHAgentStepWorkflowTest.sshAgentAvailableAfterRestart` scans the build log for the ssh-agent socket path with a regex tied to the `/tmp/ssh-XXXX/agent.<pid>` layout. On macOS the system ssh-agent creates the socket under `~/.ssh/agent/s.<id>.agent.<id>` (no `ssh-` segment, non-numeric suffix), so the pattern matched nothing and the test failed with `expected 2 but was 0`.

The socket path layout is an ssh-agent implementation detail, so the test now matches on the log label and captures whatever socket path follows. That passes on macOS and stays correct on Linux, where the old paths are just a subset of what the looser pattern accepts.

### Testing done

Reproduced on macOS Tahoe 26.5.2: the test failed before the change and passes after. The full `SSHAgentStepWorkflowTest` class (9 tests) passes with the change.

Fixes #289

### DIFF
--- a/src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
+++ b/src/test/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentStepWorkflowTest.java
@@ -197,7 +197,10 @@ class SSHAgentStepWorkflowTest extends SSHAgentBase {
 
                 j.assertBuildStatusSuccess(j.waitForCompletion(b));
 
-                Pattern pattern = Pattern.compile("(?:SSH Agent (?:before|after) restart )/.+/ssh-.+/agent.(\\d)+");
+                // The socket path layout is an ssh-agent implementation detail and varies by platform
+                // (for example macOS uses ~/.ssh/agent/s.<id>.agent.<id> rather than /tmp/ssh-XXXX/agent.<pid>),
+                // so match on the log label and capture whatever socket path follows.
+                Pattern pattern = Pattern.compile("(?:SSH Agent (?:before|after) restart )\\S+");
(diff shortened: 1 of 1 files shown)


jglick (approved):
Still unclear why this started failing in `bom` builds, which run on Linux, but :crossed_fingers:

krisstern:
I think the problem was not really related to the Linux setup for the recent BOM releases, but as some BOM maintainers like me and I think also @darinpope are using macOS this has some impact on our work while investigating this plugin. 

Thanks so much for the work @ogulcanaydogan!

PR #304, opened 2026-08-16
Model file-paths at agent computer using FilePath instead of nio.Path

HannesWell:
When using `java.nio.file.Path`, the OS and file-system of the controller is used, but the model paths in fact refer to the file-system of and files at the executing agent computer.
In case of an OS mismatch between controller and agent, this can lead to parsing errors and incorrect environments.
E.g. the Linux implementation of `java.nio.file.Path` cannot handle Windows style paths and usually throws an Exception in that case.

By using (remote) `hudson.FilePath` instead the OS/file-system of the executing agent computer is considered.

Furthermore, use the File separator of file-system at the agent computer, not the controller's.

Follow-up on
- https://github.com/jenkinsci/ssh-agent-plugin/pull/296

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
@@ -31,14 +31,13 @@ import hudson.Launcher.ProcStarter;
 import hudson.model.TaskListener;
 import hudson.slaves.WorkspaceList;
 import java.io.ByteArrayOutputStream;
-import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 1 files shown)


ogulcanaydogan:
Good catch, the controller-vs-agent filesystem mismatch is a real bug and switching to FilePath is the right fix, thanks for scoping this out as its own small PR instead of bundling it into #169, makes it much easier to review.

Heads up on timing: I have #306 open (CI running now) which touches this exact method (toSSHAgentPath) for a different bug, empty-string executable from a blank UI field crashing legacy job configs. It only adds one condition to the null-check, so once it merges this should be a trivial rebase, but you'll likely see a small conflict on that line. Happy to handle the rebase myself if that's easier for you, just say the word.

ogulcanaydogan:
Since I'm the one who created the conflict by merging #306 first, I went ahead and did the rebase myself rather than leave it on your plate. Pushed it to my own fork, not touching your branch at all: https://github.com/ogulcanaydogan/ssh-agent-plugin/tree/rebase-304-onto-master

It was a one-line conflict exactly where I expected (the null-check in toSSHAgentPath), resolved by keeping your FilePath signature with my blank-string check added in. Full suite green (50/50). Totally fine to ignore this and do your own rebase if you'd rather, just wanted to save you the trouble in case it's useful, happy to open a PR against your branch instead if that's easier than cherry-picking.

HannesWell:
> Good catch, the controller-vs-agent filesystem mismatch is a real bug and switching to FilePath is the right fix, thanks for scoping this out as its own small PR instead of bundling it into #169, makes it much easier to review.

Thanks :)

> you'll likely see a small conflict on that line. Happy to handle the rebase myself if that's easier for you, just say the word.

No problem, was easy to resolve. Just rebased and pushed the update.
But thanks for helping!

> I have #306 open (CI running now) which touches this exact method (toSSHAgentPath) for a different bug, empty-string executable from a blank UI field crashing legacy job configs

There are really a lot of things to consider :O Thanks for fixing that!

panicking (approved):

PR #305, opened 2026-08-16
Use SSH_ASKPASS_REQUIRE instead of a setting the DISPLAY variable

HannesWell:
Use environment variable `SSH_ASKPASS_REQUIRE=force` to enforce the usage of the askpass command:
- https://www.man7.org/linux/man-pages/man1/ssh-add.1.html#ENVIRONMENT

For me this makes the intend more clear and maybe could prevent any interference with an existing `DISPLAY` variable (but this is just an assumption and I didn't notice any issues).

### Testing done

Only a local execution on Windows.

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
@@ -164,7 +164,8 @@ public final class ExecRemoteAgent implements Serializable {
                 Map<String,String> env = new HashMap<>(agentEnv);
                 if (passphrase != null) {
                     env.put("SSH_PASSPHRASE", passphrase);
-                    env.put("DISPLAY", "bogus"); // just to force using SSH_ASKPASS
+                    env.put("SSH_ASKPASS_REQUIRE", "force"); // force using SSH_ASKPASS
+                    env.put("DISPLAY", "bogus"); // legacy (and backwards compatible) way to force using SSH_ASKPASS
                     env.put("SSH_ASKPASS", askpass.getRemote());
                 }
                 


ogulcanaydogan:
Nice, I like this a lot better than the old DISPLAY=bogus hack, it says what it actually means. Checked the man page and SSH_ASKPASS_REQUIRE=force is exactly right, forces askpass regardless of DISPLAY.

One thing worth double-checking before merge: SSH_ASKPASS_REQUIRE is newer than DISPLAY/SSH_ASKPASS, so an older ssh-add that doesn't recognize it would just ignore the var, and now there's no DISPLAY fallback either. Given some of the environments people run this plugin on (older RHEL/Rocky Linux boxes show up a lot in our issue tracker), I'd feel safer keeping both, i.e. leave DISPLAY=bogus in and add SSH_ASKPASS_REQUIRE=force alongside it, rather than replacing one with the other. Costs nothing on modern clients and keeps the old ones working. Happy to be wrong if you've already checked this against an older ssh-add, just wanted to flag it since your own testing note says Windows-only so far.

HannesWell:
> One thing worth double-checking before merge: SSH_ASKPASS_REQUIRE is newer than DISPLAY/SSH_ASKPASS, so an older ssh-add that doesn't recognize it would just ignore the var, and now there's no DISPLAY fallback either. Given some of the environments people run this plugin on (older RHEL/Rocky Linux boxes show up a lot in our issue tracker), I'd feel safer keeping both,

Makes sense.
From my quick research it seems that `SSH_ASKPASS_REQUIRE` is only available since about 2020. And while I hope that most people have updated their SSL impl since then, there can indeed be a chance for older systems not supporting that.
Restored the old property as suggested.

panicking (approved):

PR #312, opened 2026-08-28
fix: use delayed expansion in the Windows askpass script

ogulcanaydogan:
Fixes #311

Regression from #169 (424): before that PR the plugin only ever generated a POSIX `.sh` askpass script (correctly double-quoted since JENKINS-42093 in 2017); the Windows `.bat` branch was new code, carrying over a line (`@ECHO %SSH_PASSPHRASE%`) that had sat as a commented-out TODO since 2014 and was never actually exercised until #169 turned on real Windows execution.

`%VAR%` (immediate) expansion substitutes the passphrase before cmd.exe finishes tokenizing the rest of the line, so a passphrase containing `&`, `|`, `<`, `>`, or `^` gets re-parsed as a command separator/redirect instead of printed literally — exactly what's reported in #311. Switching to `!VAR!` (delayed expansion, via `setlocal enabledelayedexpansion`) defers the substitution until after tokenizing is done, so those characters in the passphrase's *value* are no longer re-interpreted.

### Testing done

I don't have a Windows machine to verify cmd.exe behavior directly, so I extracted the script into a `WINDOWS_ASKPASS_SCRIPT` constant and added `ExecRemoteAgentWindowsAskpassTest`, which writes it to a real `.bat` file and invokes it via `cmd.exe` with a passphrase containing `&|^<>`, asserting the captured stdout matches exactly. It's gated on `Functions.isWindows()` so it skips everywhere except the `windows-21` CI job, where it should give a real answer rather than my own reasoning about cmd.exe parsing rules. Full suite passes locally (60 tests, 1 skipped for the same reason).

Known limitation: a literal `!` in the passphrase would still be misinterpreted as the start of another delayed-expansion reference (a well-known tradeoff of this technique). That's a narrower, silent-corruption failure mode versus the current 100%-broken-on-`&` one, and I didn't see an approach that fixes both without adding a new dependency (e.g. shelling out to PowerShell) to what's otherwise a plain batch script.

Issue #311, opened 2026-08-27
ssh-add fails only on Windows when SSH key passphrase contains & -- introduced in ssh-agent plugin 427.v5818150c4b_c6 (latest)


sylvainchagnaud:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
  Jenkins: 2.568.1
OS: Linux - 6.8.0-1064-azure
Java: 21.0.11 - Eclipse Adoptium (OpenJDK 64-Bit Server VM)
---
Parameterized-Remote-Trigger:3.2.2
ace-editor:1.1
active-directory:2.967.vf4e254546a_b_3
analysis-model-api:14.14.0-1004.vf53c9efb_f455
ansicolor:536.v13fa_b_860c267
ant:520.vd082ecfb_16a_9
antisamy-markup-formatter:173.v680e3a_b_69ff3
apache-httpcomponents-client-4-api:4.5.14-269.vfa_2321039a_83
apache-httpcomponents-client-5-api:5.6.2-199.vc3c04d033fcc
artifactory:4.0.8
asm-api:9.10.1-216.va_9256d3b_844b_
authentication-tokens:1.144.v5ff4a_5ec5c33
authorize-project:531.vf8761dfd75d2
azure-ad:699.v6dd03c77b_981
azure-credentials:426.v5627e6a_4c7a_7
azure-sdk:284.vb_4128d6c7561
badge:2.562.vc3c0b_6925377
basic-branch-build-strategies:317.v85b_331d6cc42
blueocean:1.27.25
blueocean-autofavorite:1.2.5
blueocean-bitbucket-pipeline:1.27.25
blueocean-commons:1.27.25
blueocean-config:1.27.25
blueocean-core-js:1.27.25
blueocean-dashboard:1.27.25
blueocean-display-url:2.4.4
blueocean-events:1.27.25
blueocean-git-pipeline:1.27.25
blueocean-github-pipeline:1.27.25
blueocean-i18n:1.27.25
blueocean-jira:1.27.25
blueocean-jwt:1.27.25
blueocean-personalization:1.27.25
blueocean-pipeline-api-impl:1.27.25
blueocean-pipeline-editor:1.27.25
blueocean-pipeline-scm-api:1.27.25
blueocean-rest:1.27.25
blueocean-rest-impl:1.27.25
blueocean-web:1.27.25
bootstrap4-api:4.6.0-6
[...]

PR #314, opened 2026-08-30
Read passphrase from env variable again in askpass script for Windows

HannesWell:
Further refine the askpass bat script used on Windows introduced in
- https://github.com/jenkinsci/ssh-agent-plugin/pull/312

to echo the passphrase environment variable reliably even if it contains exclamation marks, any other special character or the Windows bat help command '/?'.

Additionally extend the `ExecRemoteAgentWindowsAskpassTest` to test more difficult passphrases.

This also reverts https://github.com/jenkinsci/ssh-agent-plugin/pull/313 and avoids storing decrypted passphrases in plain text in temporary files and refines the fix for
- https://github.com/jenkinsci/ssh-agent-plugin/issues/311

### Testing done

Local testing of the bat script on my Windows computer and the `ExecRemoteAgentWindowsAskpassTest` tests are extended.

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
@@ -210,15 +210,15 @@ public final class ExecRemoteAgent implements Serializable {
         try {
             keyFile.chmod(0600);
 
-            AskpassFiles askpass = passphrase != null ? createAskpassScript(temp, passphrase) : null;
+            FilePath askpass = null;
             try {
-
                 Map<String,String> env = new HashMap<>(agentEnv);
                 if (passphrase != null) {
+                    askpass = createAskpassScript(temp, isWindowsAgent);
                     env.put("SSH_PASSPHRASE", passphrase);
                     env.put("SSH_ASKPASS_REQUIRE", "force"); // force using SSH_ASKPASS
                     env.put("DISPLAY", "bogus"); // legacy (and backwards compatible) way to force using SSH_ASKPASS
-                    env.put("SSH_ASKPASS", askpass.script().getRemote());
+                    env.put("SSH_ASKPASS", askpass.getRemote());
                 }
 
                 // as the next command is in quiet mode, we just add a message to the log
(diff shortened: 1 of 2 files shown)


src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
```diff
@@ -332,56 +325,35 @@ static String getAgentValue(String agentOutput, String envVar) throws AbortExcep
...
+            // Delayed expansion (!VAR! instead of %VAR%) defers substitution of SSH_PASSPHRASE until
+            // after cmd.exe has already tokenized the line, so characters in the passphrase such as
+            // & or | are no longer re-parsed as command separators.
+            // See https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/setlocal#parameters
+            // Using echo( ensures empty passphrases or those containing ! work too
+            askpass = temp.createTextTempFile("askpass_", ".bat", """
+                    @SetLocal EnableDelayedExpansion
+                    @ECHO(!SSH_PASSPHRASE!
+                    @start /b cmd /c del "%~f0" & exit /b
+                    """, !pathContainsSpaces);
```
HannesWell, line 346: The critical difference to before is using `ECHO(` instead of just `ECHO `.
As per my research this is considered the most robust way to echo a variable verbatim.
It also works if `SSH_PASSPHRASE` is empty or the string `/?`.
In general `echo /?` prints the help of the echo command (similar to --help for many Unix commands).

Furthermore turning of the command echo for the start command allows to make tests more strict.

HannesWell:
@ogulcanaydogan could you please review this and if you think it's suitable submit and release it?

ogulcanaydogan:
`ECHO(` is a nice trick, didn't know that one. Genuinely better than what I had in #313: no plaintext passphrase touching disk, and it catches two cases I'd missed entirely (empty passphrase hitting "ECHO is on.", and a literal `/?` triggering ECHO's help text) that would have been broken either way. Tested locally and CI's green including windows-21. Happy to have this replace #313.

HannesWell:
Thanks for reviewing and submitting this!

> it catches two cases I'd missed entirely (empty passphrase hitting "ECHO is on.", and a literal `/?` triggering ECHO's help text) that would have been broken either way.

To be fair, I'm not even sure if an empty passphrase would be possible with ssh, because if one generates the keys, inputting an 'empty' string usually means no passphrase.
And just `/?` would be a pretty weak passphrase. But I think it's better to have it covered.

ogulcanaydogan:
Fair point on both, agreed neither is a realistic real-world passphrase. Cheap insurance either way though, and it made a nice property to assert in the test.

PR #316, opened 2026-09-01
Extend PATH with <git-home>\usr\bin if git's ssh-agent is used

HannesWell:
In general the 'UNIX' ssh-agent provided by git-for-windows, which is used by this plugin on Windows agents (if no other is specified) works together with the ssh tools provided natively by Windows. However the native Windows ssh tools have multiple quirks and do not always work without problem together with the 'UNIX' ssh tools. In order to further reduce the friction when using this plugin on Windows, prepend the `<git-home>\usr\bin` directory to the PATH environment variable within the sshagent's code block. With that the ssh tools (like ssh, scp, ssh-agent, ...) provided by git-for-windows are used e.g. in scripts within the sshagent block.

But the PATH environment variable is only extended when running on Windows and if this plugin selects the sshagent automatically, i.e. if no executable is specified.

This was notcied after testing https://github.com/jenkinsci/ssh-agent-plugin/pull/169 in detail in production, where I for example encountered issues like
- https://serverfault.com/questions/994646/ssh-on-windows-corrupted-mac-on-input

In the current state it often unfortunately does not 'just' work.
On the other hand adding `C:\Program Files\Git\usr\bin` to PATH globally also has some drawbacks, since for example the UNIX `tar.exe` cannot handle zip files, while the `tar.exe` provided on Windows natively does.

I've also considered to add a new property named `addUnixToolsForWindowsToPATH` allowing a user to disable the automated PATH extension. But one could achieve the same intermediate state (i.e. using `C:\Program Files\Git\usr\bin\ssh-agent.exe` without extending PATH), by setting the `executable` property to `C:\Program Files\Git\usr\bin\ssh-agent.exe`.

### Testing done

Local testing on a Windows computer with and without this change

### Submitter checklist

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
@@ -53,6 +53,7 @@ import java.util.regex.Pattern;
 public final class ExecRemoteAgent implements Serializable {
     private static final String AuthSocketVar = "SSH_AUTH_SOCK";
     private static final String AgentPidVar = "SSH_AGENT_PID";
+    private static final String SSHAGENT_USR_BIN_PATH_EXTENSION = "PATH+SSH_AGENT_USR_BIN";
 
     /** Default timeout, in minutes, applied to the agent commands when none is configured. */
     public static final int DEFAULT_TIMEOUT_MINUTES = 1;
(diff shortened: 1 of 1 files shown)


HannesWell:
Thank you @ogulcanaydogan for reviewing and submitting.
Could you also be so kind to create a new release with this included? Thanks in advance.

HannesWell:
@ogulcanaydogan could you please publish a release with this change included? That would be really helpful. Thanks in advance.

ogulcanaydogan:
Published as 432.v122f1207432b_

PR #317, opened 2026-09-06
fix: honor a custom PATH and pass step/build env to ssh-agent, ssh-add

ogulcanaydogan:
### Summary

`SSHAgentStepExecution` and `SSHAgentBuildWrapper` never passed the step's/build's environment through to `ExecRemoteAgent`. As a result:

- A custom `PATH` set for the block (e.g. via `withEnv`) was not honored when locating the default `ssh-agent` executable, since a bare command name is resolved against the launcher process' own ambient PATH, not the environment passed to the launched process.
- Other environment variables set for the block never reached the `ssh-agent`/`ssh-add` commands either.

This is the underlying issue described in #227 (originally reported against the exec launcher not considering the step's environment).

### Change

- Thread the context `EnvVars` into `ExecRemoteAgent` (new constructor overload; the old one delegates with an empty `EnvVars` for compatibility) and pass them on every launch of `ssh-agent`/`ssh-add`/`ssh-agent -k`.
- If the default `ssh-agent` lookup still fails, retry once by resolving `ssh-agent` against the context's `PATH` explicitly (see `findOnPath`). This is only a fallback for the non-Windows, no-explicit-`executable` case, so default behavior for users who don't override `PATH` is unchanged (verified via the existing test suite, including `SSHAgentBuildWrapperTest`, which failed with an eagerly-resolved approach since the ambient `PATH` on any real machine already contains `ssh-agent`).

### Testing

- Added unit tests for `findOnPath` (locates the executable on a `PATH` entry, searches multiple colon-separated directories in order, returns `null` when unset/blank or not found).
- Ran the full existing test suite (66 tests) plus `mvn verify` (SpotBugs/Checkstyle) locally, all green.

### DIFF
--- a/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
+++ b/src/main/java/com/cloudbees/jenkins/plugins/sshagent/SSHAgentBuildWrapper.java
@@ -31,6 +31,7 @@ import com.cloudbees.plugins.credentials.common.StandardUsernameCredentials;
 import com.cloudbees.plugins.credentials.common.StandardUsernameListBoxModel;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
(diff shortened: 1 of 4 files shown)


ogulcanaydogan:
Closing per the discussion on #227 — the original PATH problem is already solved by git-ssh-agent auto-detection, and I couldn't point to a concrete use case for the rest of this beyond flexibility for its own sake.

Also noticed #318 touches the same `SSHAgentStepExecution.java` for a much more substantial fix (the parallel-pipeline serialization issue), so keeping this open would just create merge friction for no real benefit. Happy to revisit if a concrete env-var use case turns up later.

PR #318, opened 2026-09-14
Honey I shrunk the queue: Unblock parallel pipelines

samrocketman:
Part of my "honey I shrunk the queue" patch series where I solve a significant performance problem whose fix is required across multiple plugins.  Search for the same phrase on the Jenkins org to see similar patches related to this.

Why this matters?
-----------------

- This unblocks pipelines serializing commands run from within `sshagent` steps.
- Fixing this will positively impact declarative pipeline performance when `sshagent` steps are used in parallel contexts.

`sshagent(...)` blocks the top-level CPS VM thread and causes a parallel pipeline to go very slow (serially).

- Before this patch: my production environment with a simple "echo hello" run in parallel takes 10-20 minutes to run.
- After this patch: 100 nodes execute in parallel in under 30 seconds.

<details><summary>The queue has shrunk like the kids</summary>

---

From the movie: Honey I Shrunk the Kids

image

</details>

---

Sample Jenkinsfile
------------------

If you're willing to provision 100 nodes of compute to test you can replicate the forced serialization with the following scripted pipeline `Jenkinsfile`.

```groovy
Map parallelAgents = [failFast: false]
for (int i = 1; i <= 100; i++) {
    String agentNum = "Agent ${agentNum}"
    parallelAgents["Agent ${agentNum}"] = { ->
        stage(agentNum) {
            echo 'before node'
            node('some-label') {
                echo 'before docker'
                docker.image('alpine').inside('--init') {
                    echo 'before sshagent'
                    sshagent(['some-credentials']) {
                        echo 'before sh'
                        sh "echo hello from ${agentNum}"
                    }
                }
            }
        }
    }
}
parallel(parallelAgents)
```

AI Analysis
-----------

You may find my AI analysis at https://github.com/samrocketman/jenkins-ai-analysis/tree/main/honey-i-shrunk-the-queue

Testing done
------------

Tested in my staging Jenkins infrastructure while I was pressure testing https://github.com/jenkinsci/ec2-plugin/pull/2036

Related issues
--------------

None.  This feature is fixing a problem that likely contributed to the cause of closed issues #277 and #268.  AI analysis suggests #253 could be addressed but I don't see how; the real fix would be something else.

Submitter checklist
-------------------


samrocketman:
I commit to post-merge support for this if it causes issues. Though I think the risk is low since I tested with a fairly tall scale (my test pipeline is more thorough than the sample reproducing Jenkinsfile).

ogulcanaydogan (approved):
Solid fix. `GeneralNonBlockingStepExecution` + `run(this::doStart)` correctly moves the blocking launch off the CPS VM thread, and `stopAgentAsync` restoring the auth via `ACL.as2(auth)` before running on `Computer.threadPoolForRemoting` is the right way to keep the teardown authorized once it's off-thread. The volatile fields make sense given `agent`/`username`/etc. are now written on a background thread and read from the CPS thread. Regression test is exactly the right shape — a stand-in `ssh-agent` that blocks until released, with a `parallel` sibling branch proving the CPS VM thread isn't stuck.

Real-world numbers in the description (10-20min → 30s for 100 parallel nodes) plus green CI on both platforms. Merging.

PR #319, opened 2026-09-20
fix: support Cygwin ssh-agent path layout on Windows

mahdipub:
## Problem

Since v424 the plugin introduced official Windows support by searching for `ssh-agent.exe` under the Git installation using this hardcoded sub-path:

```java
private static final String GIT_SSH_AGENT_PATH_WINDOWS = "usr\\bin\\ssh-agent.exe";
```

This works for **Git-for-Windows** installs (e.g. `C:\Program Files\Git\usr\bin\ssh-agent.exe`), but breaks **Cygwin**-based nodes, where the binary lives at `<cygwin-root>\bin\ssh-agent.exe` — **without the `usr\` prefix**.

When Cygwin's `git.exe` is on `PATH`, the plugin resolves:
```
where git  =>  C:\cygwin64\bin\git.exe
ssh-agent  =>  C:\cygwin64\usr\bin\ssh-agent.exe   (does not exist)
```
and throws:
```
java.lang.IllegalStateException: Executing with default ssh-agent on Windows is not supported
and an alternative implementation from a git installation is not available.
    at ExecRemoteAgent.lambda$searchSSHAgentExeForWindows$5(ExecRemoteAgent.java:129)
```

This broke **100% of Windows jobs** on our Jenkins instance immediately after the TaaS-managed upgrade to v430. The regression affects any node using Cygwin as the Git provider.

We confirmed this on a Windows Server 2025 x86-64 node:

| Probe | Result |
|---|---|
| `where git` | `C:\cygwin64\bin\git.exe` |
| `C:\cygwin64\usr\bin` exists? | **NO** |
| `C:\cygwin64\bin\ssh-agent.exe` exists? | **YES** |
| Plugin result | `IllegalStateException` |

## Fix

Probe both candidate sub-paths in order, returning the first that exists:

1. `usr\bin\ssh-agent.exe` — Git-for-Windows layout (existing behaviour, unchanged)
2. `bin\ssh-agent.exe` — Cygwin layout (new fallback)

The change is confined to `extractGitSSHAgentExe()` and is fully backwards-compatible — Git-for-Windows nodes are unaffected because the `usr\bin` candidate is tried first.

## Tests

Four new unit tests added to `ExecRemoteAgentTest`:

| Test | What it verifies |
|---|---|
| `findsSSHAgentUnderUsrBinForGitForWindows` | Git-for-Windows layout still works (regression guard) |
| `findsSSHAgentUnderBinForCygwin` | Cygwin layout now found via new `bin/` fallback |
| `returnsEmptyWhenSSHAgentNotFound` | Empty result when neither path exists |
| `findsSSHAgentWhenGivenGitExePathInsteadOfGitHome` | Path stripping + Cygwin fallback when given a `git.exe` path directly (Windows-only, matches `where git` output) |

All pass on Linux/macOS (Windows-only test skipped via `@EnabledOnOs(OS.WINDOWS)`).

## Checklist
- [x] Topic/feature branch (not main)
- [x] PR title describes the changelog entry
- [x] Description of the change
- [x] Tests covering both fix and regression
- [x] Linked to related upstream work (#169, #307)


mahdipub:
Closing in favour of a cleaner replacement PR that matches the project's contribution style.

PR #320, opened 2026-09-20
fix: support Cygwin ssh-agent path layout on Windows

mahdipub:
On Windows, the plugin resolves `ssh-agent.exe` by appending `usr\bin\ssh-agent.exe` to the git home it finds via `where git`. This works for Git-for-Windows, but Cygwin places the binary at `<cygwin-root>\bin\ssh-agent.exe` — no `usr\` prefix — causing an `IllegalStateException` on every Windows job when Cygwin is used as the git provider.

The fix probes both candidate paths in order: `usr\bin\ssh-agent.exe` first (Git-for-Windows, existing behaviour unchanged), then `bin\ssh-agent.exe` as a Cygwin fallback.

Follow-up on
- https://github.com/jenkinsci/ssh-agent-plugin/pull/169
- https://github.com/jenkinsci/ssh-agent-plugin/pull/307

### Testing done

Four new unit tests added to `ExecRemoteAgentTest` covering: Git-for-Windows layout (regression guard), Cygwin layout (the fix), no ssh-agent present (empty result), and git.exe path input with path stripping (Windows-only via `@EnabledOnOs`). All pass.

### Submitter checklist


mahdipub:
@ogulcanaydogan @HannesWell — would appreciate a review when you have a moment.

This fixes the Cygwin path regression introduced in #169. `HannesWell` — since you authored the original Windows support, your eye on this would be especially valuable.

mahdipub:
@HannesWell do you mind take a look at this PR? It tries to cover the issue introduced by PR #169.
CC: @ogulcanaydogan for merge please

andrew-m-leonard (approved):
Looks good thanks @mahdipub we are seeing this issue at Adoptium as well

src/main/java/com/cloudbees/jenkins/plugins/sshagent/exec/ExecRemoteAgent.java
```diff
@@ -137,18 +137,25 @@ private FilePath searchSSHAgentExeForWindows(Launcher launcher, TaskListener lis
...
             if (GIT_EXE_PATH.matcher(path).find()) {
                 git = git.map(FilePath::getParent).map(FilePath::getParent);
             }
-            Optional<FilePath> sshAgentExe = git.map(p -> p.child(GIT_SSH_AGENT_PATH_WINDOWS));
-            if (sshAgentExe.isPresent() && sshAgentExe.get().exists()) {
-                return sshAgentExe;
+            // Probe candidate sub-paths: Git-for-Windows first, then Cygwin layout.
+            for (String candidatePath : GIT_SSH_AGENT_CANDIDATE_PATHS_WINDOWS) {
+                final String candidate = candidatePath;
+                Optional<FilePath> sshAgentExe = git.map(p -> p.child(candidate));
```
HannesWell, line 155: I'd say the comment is not necessary since the code is clear enough and it would be another place that needs updates in case the list is extended.
Also the separate candidate variable is not necessary.
```suggestion
            for (String agentPath : GIT_SSH_AGENT_CANDIDATE_PATHS_WINDOWS) {
                Optional<FilePath> sshAgentExe = git.map(p -> p.child(agentPath));
```

HannesWell (commented):
This looks reasonable for me. Thanks for this contribution.

IMO there is just one nit-pick, but eventually @ogulcanaydogan has to decide anyway.

